Issue #7, opened 2019-11-05
Use more descriptive assertions


ajaust:
Hi,

I think it would make sense to have more descriptive assertions, especially with the changes in the new Python bindings (`python_future`). A lot of things changed in the API making it likely that things break in user code. I was a bit surprised that the new bindings not only change the number of arguments a preCICE function takes (`set_mesh_vertices` takes 3 arguments instead of 4 in the C++ interface), but also it expects different argument types/shapes. 

In the old bindings `set_mesh_vertices` would accept a list of coordinates where it was basically a 1D list/array of length `numPoints*dimension`. The data passed to tthe function did **not** have to be an numpy array. In `precice_future` it worked for me only when I use 2D numpy.array with `numPoints` entries that have the coordinates (`dimension`).

I started with using an older code with the new interface that used a 1D list (no numpy) of coordinates. This was caught by the exception [in line 135](https://github.com/precice/precice/blob/v1.6.1/src/precice/bindings/python_future/precice_future.pyx#L135), but the exception was not really helpful as there was no explanation what I did wrong. From the documentation it is also not obvious what type/dimension the input arguments should have. 

I would suggest to replace the line
```python
assert(dimensions == self.get_dimensions())
```
by something like
```python
assert(dimensions == self.get_dimensions(),"Descriptive message what failed and why (what did the binding expect here?)")
```

This holds true for the other assertions as well. At least for me as a user it was not immediately obvious why it failed and what the binding actually expects. This is also somehow related to the documentation of the bindings as discussed in precice/precice#450 and precice/python-bindings#3.

Thank you!

PR #9, opened 2019-12-03
Improved error messages for bindings

BenjaminRodenberg:
Migrated https://github.com/precice/precice/pull/547 here.

Closes https://github.com/precice/python-bindings/issues/74, closes https://github.com/precice/python-bindings/issues/7


BenjaminRueth:
The implementation of the checks looks good. However, the checks are actually not used. I assume that the relevant code has been removed within a merge or similar?

BenjaminRueth:
> The implementation of the checks looks good. However, the checks are actually not used. I assume that the relevant code has been removed within a merge or similar?

https://github.com/precice/precice/pull/547 contains the relevant parts. I will manually move them to this PR.

IshaanDesai:
Need to provide a good error message for: https://github.com/precice/python-bindings/issues/74

BenjaminRodenberg:
#74 gives an example how we can deal with certain kinds of errors and how we can also test that under certain conditions the correct kind of error is thrown.

BenjaminRodenberg:
Some more things that we should check (not covered by #74):

* If we expect an id it should be an int
* If we expect ids it should be an array_like of ints
* If we expect a single value it should not be an array_like (non-block operations)

Probably there is much more, but this is what currently comes to my mind.

IshaanDesai:
At the moment only the existing assertion statements have been made more descriptive. There is scope for adding further checks and this is up for discussion

cyprecice/cyprecice.pyx
```diff
@@ -985,12 +988,11 @@ cdef class Interface:
         if len(values) == 0:
             size = 0
 
-
         cdef np.ndarray[int, ndim=1] _vertex_ids = np.ascontiguousarray(vertex_ids, dtype=np.int32)
         cdef np.ndarray[double, ndim=1] _values = np.ascontiguousarray(values, dtype=np.double)
         
-        assert(_values.size == size)
-        assert(_vertex_ids.size == size)
+        assert _values.size == size, "Scalar data is not provided for all vertices in write_block_scalar_data. Check length of input data provided"
```
BenjaminRodenberg, line 994:
```suggestion
        assert _values.size == size, "Scalar data is not provided for all vertices in write_block_scalar_data. Check length of input data provided. Provided size: {}, Expected size: {}".format(_values.size, size)
```
I think it would be helpful to also provide the values that the assertion compares.

BenjaminRodenberg (commented):
Generally helpful error messages. If you want, you can also provide the values that go into the assertion. I'm giving an example below. But this is optional.

IshaanDesai (approved):
Only the following test fails: https://github.com/precice/python-bindings/pull/9/checks?check_run_id=2176968006
Rest from my side this PR is ready to merge.

BenjaminRodenberg:
> Only the following test fails: https://github.com/precice/python-bindings/pull/9/checks?check_run_id=2176968006
> Rest from my side this PR is ready to merge.

I think the failing test should have been fixed on develop already. Just merged via 1d76a09 and hope it's working now.

Issue #11, opened 2019-12-06
Use single step setup for python bindings


BenjaminRueth:
See https://github.com/precice/precice/issues/366

PR #13, opened 2020-01-14
Moving configure call into interface initialization

IshaanDesai:
With reference to https://github.com/precice/precice/issues/366, the configuration call is transferred into the interface initialization call 
Closes https://github.com/precice/python-bindings/issues/11

### DIFF
--- a/precice.pyx
+++ b/precice.pyx
@@ -29,17 +29,14 @@ cdef class Interface:
     # construction and configuration
     # constructor
 
-    def __cinit__ (self, solver_name, solver_process_index, solver_process_size):
+    def __cinit__ (self, solver_name, configuration_file_name, solver_process_index, solver_process_size):
         self.thisptr = new SolverInterface.SolverInterface (convert(solver_name), solver_process_index, solver_process_size)
+        self.thisptr.configure (convert(configuration_file_name))
         pass
 
     # destructor
     def __dealloc__ (self):
         del self.thisptr
-
-    # configure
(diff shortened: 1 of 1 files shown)


precice.pyx
```diff
@@ -29,18 +29,15 @@ cdef class Interface:
     # construction and configuration
     # constructor
 
-    def __cinit__ (self, solver_name, solver_process_index, solver_process_size):
+    def __cinit__ (self, solver_name, configuration_file_name, solver_process_index, solver_process_size):
         self.thisptr = new SolverInterface.SolverInterface (convert(solver_name), solver_process_index, solver_process_size)
+        self.thisptr.configure (convert(configuration_file_name))
```
BenjaminRueth: See https://github.com/precice/precice/issues/366#issuecomment-574218275. As soon as the single step setup in C++ is ready the lines above should be changed to

```
self.thisptr = new SolverInterface.SolverInterface (convert(solver_name), convert(configuration_file_name), solver_process_index, solver_process_size)
```

IshaanDesai: Yes, I agree that regarding the single step initialization the order of merging would be:
1. Change the C++ interface
2. Change python-bindings
3. Change tutorials, tests and every other possible python related code piece

BenjaminRueth: I realized that waiting for the C++ interface to be ready will delay this PR (+ the many dependent issues) unnecessarily, since it is relatively easy to change the way how the python bindings access the C++ interface at a later point in time "under the hood".

Issue #3, opened 2019-06-14
Proper documentation for new python bindings


IshaanDesai:
The [refactored](https://github.com/precice/precice/tree/develop/src/precice/bindings/python) python bindings and underlying [new](https://github.com/precice/precice/tree/develop/src/precice/bindings/python_future) bindings need documentation explaining the API functions and also some basic explanation of cython working

IshaanDesai:
Related PR is now in the new `python-bindings` repo: https://github.com/precice/python-bindings/pull/10

PR #33, opened 2020-02-06
docstring documentation for python-bindings

IshaanDesai:
Closes https://github.com/precice/python-bindings/issues/3


precice.pyx
```diff
@@ -38,84 +58,328 @@ cdef class Interface:
...
+
+        Returns:
+            id_array (array of ints): Set of IDs
+        """
         return self.thisptr.getMeshIDs ()
 
-    # returns a handle to a created mesh
+
     def get_mesh_handle(self, mesh_name):
+        """
```
BenjaminRueth: Add note in documentation that method is not available yet.

precice.pyx
```diff
@@ -124,12 +388,39 @@ cdef class Interface:
...
-    # creates multiple mesh vertices
     def set_mesh_vertices (self, mesh_id, positions):
+        """
+        Creates multiple mesh vertices
+
+        Parameters:
+            mesh_id (int): ID of the mesh to add the vertices to.
+            positions (pointer): a pointer to the coordinates of the vertices
+                                 The 2D-format is (d0x, d0y, d1x, d1y, ..., dnx, dny)
+                                 The 3D-format is (d0x, d0y, d0z, d1x, d1y, d1z, ..., dnx, dny, dnz)
```
BenjaminRueth: This has to be updated to the format that the python bindings are using. I.e a `numpy` array of shape `N x D`. Consider giving an example, as well. Compare to `help(np.array)`, where at the bottom of the documentation examples are provided.

precice.pyx
```diff
@@ -139,16 +430,47 @@ cdef class Interface:
...
         return _ids
 
-    # get vertex positions for multiple vertex ids from a given mesh
     def get_mesh_vertices(self, mesh_id, ids):
+        """
+        Get vertex positions for multiple vertex ids from a given mesh
+
+        Parameters:
+            mesh_id (int): ID of the mesh to read the vertices from.
+            ids (array): The ids of the vertices to lookup
```
BenjaminRueth: `numpy.array` of shape `N`

precice.pyx
```diff
@@ -139,16 +430,47 @@ cdef class Interface:
...
+
+        Previous calls:
+            count of available elements at positions matches the configured dimension * size
+            count of available elements at ids matches size
+
+        Returns:
+            positions (pointer): a pointer to memory to write the coordinates to
+                                 The 2D-format is (d0x, d0y, d1x, d1y, ..., dnx, dny)
+                                 The 3D-format is (d0x, d0y, d0z, d1x, d1y, d1z, ..., dnx, dny, dnz)
```
BenjaminRueth: See above. This is the C++ format.

precice.pyx
```diff
@@ -139,16 +430,47 @@ cdef class Interface:
...
     def get_mesh_vertex_ids_from_positions (self, mesh_id, positions):
+        """
+        Gets mesh vertex IDs from positions.
+        prefer to reuse the IDs returned from calls to set_mesh_vertex() and set_mesh_vertices().
+
+        Parameters:
+            mesh_id (int): ID of the mesh to retrieve positions from
+            positions (n dims array): Positions to find ids for.
+                                      The 2D-format is (d0x, d0y, d1x, d1y, ..., dnx, dny)
+                                      The 3D-format is (d0x, d0y, d0z, d1x, d1y, d1z, ..., dnx, dny, dnz)
```
BenjaminRueth: See above. This is the C++ format.

precice.pyx
```diff
@@ -158,41 +480,168 @@ cdef class Interface:
...
+        """
+        Writes vector data given as block. This function writes values of specified vertices to a dataID.
+        Values are provided as a block of continuous memory. valueIndices contains the indices of the vertices
+        The 2D-format of values is (d0x, d0y, d1x, d1y, ..., dnx, dny)
+        The 3D-format of values is (d0x, d0y, d0z, d1x, d1y, d1z, ..., dnx, dny, dnz)
+
+        Parameters:
+            data_id (int): ID to write to.
+            value_indices (array): Indices of the vertices.
+            values (pointer): a pointer to the vector values.
```
BenjaminRueth: See above. This is the C++ format.

precice.pyx
```diff
@@ -158,41 +480,168 @@ cdef class Interface:
...
+            A mapping from from_mesh_id was configured.
+        """
         self.thisptr.mapWriteDataFrom (from_mesh_id)
 
     def write_block_vector_data (self, data_id, value_indices, values):
+        """
+        Writes vector data given as block. This function writes values of specified vertices to a dataID.
+        Values are provided as a block of continuous memory. valueIndices contains the indices of the vertices
+        The 2D-format of values is (d0x, d0y, d1x, d1y, ..., dnx, dny)
+        The 3D-format of values is (d0x, d0y, d0z, d1x, d1y, d1z, ..., dnx, dny, dnz)
```
BenjaminRueth: See above. This is the C++ format.

precice.pyx
```diff
@@ -204,6 +653,24 @@ cdef class Interface:
...
+        """
+        Wrties vector data to a vertex. This function writes a value of a specified vertex to a dataID.
+        Values are provided as a block of continuous memory.
+        The 2D-format of value is (x, y)
+        The 3D-format of value is (x, y, z)
+
+        Parameters:
+            data_id (int): ID to write to.
+            value_index (int): Index of the vertex.
+            value (pointer): pointer to the vector value.
```
BenjaminRueth: See above. This is the C++ format.

precice.pyx
```diff
@@ -212,16 +679,70 @@ cdef class Interface:
         self.thisptr.writeVectorData (data_id, value_index, <const double*>_value.data)
 
     def write_block_scalar_data (self, data_id, value_indices, values):
+        """
+        Writes scalar data given as a block. This function writes values of specified vertices to a dataID.
+        Values are provided as a block of continuous memory. valueIndices contains the indices of the vertices
+
+        Parameters:
+            data_id (int): ID to write to.
+            value_indices (array): Indices of the vertices.
```
BenjaminRueth: See above. This is the C++ format.

precice.pyx
```diff
@@ -158,41 +480,168 @@ cdef class Interface:
...
+            values (pointer): a pointer to the vector values.
+
+        Previous calls:
+            count of available elements at values matches the configured dimension * size
+            count of available elements at valueIndices matches the given size
+            initialize() has been called
+
+        Refer:
+            Interface::set_mesh_vertex()
```
BenjaminRueth: Why do we refer to `set_mesh_vertex()`?

IshaanDesai: I actually do not have a good explanation for this. It was there before so I left it there. I also see no relevance.

precice.pyx
IshaanDesai: All the block commands and also the relevant `set` and `get` commands have the updated `numpy.array([N, D])` documented along with examples.

precice.pyx
IshaanDesai: This change was also implemented in other locations where it was relevant to mention that not just an `array` but a `numpy.array` of shape `N` is used

precice.pyx
BenjaminRueth: I think it comes from the doxygen documentation of the c++ interface (see https://github.com/precice/precice/blob/6630c77a76b51503a09cd4c2f8425d976aa46bb9/src/precice/SolverInterface.hpp#L598). The main purpose here is to provide a link in order to be able to quickly navigate. 

If we use `help(...)` for showing the documentation this feature does not work. I am not sure how `readthedocs` deals with this keyword.

For now I would suggest to remove all the `Refer:` sections, since without links they do not really help. As soon as we are using `readthedocs` we might consider reintroducing them.

precice.pyx
```diff
@@ -212,16 +743,85 @@ cdef class Interface:
...
 
     def read_block_vector_data (self, data_id, value_indices):
+        """
+        Reads vector data into a provided block. This function reads values of specified vertices
+        from a dataID. Values are read into a block of continuous memory. valueIndices contains
+        the indices of the vertices.
+        Return values are stored in a numpy array [N x D] where N = number of vertices
+        and D = dimensions of geometry
+
+        Examples:
```
BenjaminRueth: According to https://numpydoc.readthedocs.io/en/latest/format.html#sections, the Examples go after the Parameters.

precice.pyx
```diff
@@ -139,16 +499,97 @@ cdef class Interface:
...
+        >>> mesh_id = "my_mesh"
+        >>> ids = [1, 2, 3, 4, 5]
+        >>> positions = precice.get_mesh_vertices(mesh_id, ids)
+        >>> positions.shape
+        (5, 2)
+
+        Return data structure for a 3D problem with 5 vertices:
+        >>> mesh_id = "my_mesh"
+        >>> ids = [1, 2, 3, 4, 5]
+        >>> positions = precice.get_mesh_vertices(mesh_d, ids)
```
BenjaminRueth:
```suggestion
        >>> positions = precice.get_mesh_vertices(mesh_id, ids)
```

Issue #8, opened 2019-12-03
How do we want to deal with the mpi4py dependency?


BenjaminRueth:
Continue closed PR https://github.com/precice/precice/pull/312 here.

BenjaminRueth:
I did some history research:

In https://github.com/precice/precice/pull/299 we observed that `mpi4py` was missing in the python solverdummy. In https://github.com/precice/precice/pull/316, we decided to move it directly into the bindings and remove it from the solverdummy.

uekerman:
There is also https://github.com/precice/precice/issues/311

BenjaminRueth:
I looked a bit more into history and found the original reason for adding the statement `from mpi4py import MPI` [here](https://github.com/precice/precice/pull/299#issuecomment-469421697).

I would suggest to first reproduce this kind of behaviour in a test in order to make sure that it's worth all the trouble.

BenjaminRueth:
In the following I will provide a description how to provoke the "mpi4py" error mentioned in https://github.com/precice/precice/pull/299#issuecomment-469421697

## Preparations

### install preCICE
Use preCICE revision https://github.com/precice/precice/commit/9f778290416416255fc73a495e962def301648b0

Build and install preCICE via
```
mkdir build
cd build
cmake -DBUILD_SHARD_LIBS=ON -DPETSC=OFF -DMPI=<ON|OFF>.
make -j4
sudo make install
```

### Install python bindings

Use python bindings revision https://github.com/precice/python-bindings/commit/7ddf2894644bb596e3ddbf772059ed98ab61b5ed and remove [this line](https://github.com/precice/python-bindings/blob/7ddf2894644bb596e3ddbf772059ed98ab61b5ed/precice.pxd#L10)

Install bindings via `pip3 install --user .`

### Run solverdummy

Navigate to `cd precice/tools/solverdummies`

Run `python3 python3 python/solverdummy.py precice-config.xml SolverOne MeshOne` and `python3 python/solverdummy.py precice-config.xml SolverTwo MeshTwo`

### Outcome

1. If `-DMPI=OFF`, everything works as expected
2. If `-DMPI=ON`, we get the following error:

```
~/precice/tools/solverdummies$ python3 python/solverdummy.py precice-config.xml SolverOne MeshOne
[2020-01-16 21:18:51.319269] [0x00007f78ae21db80] [trace]   Entering operator()
[2020-01-16 21:18:51.319303] [0x00007f78ae21db80] [debug]   Initialize MPI
[...]

PR #36, opened 2020-02-10
Remove MPI from python bindings

BenjaminRueth:
The way how the python bindings are using MPI makes the code more verbose, complicated compilation and leads to a confusing dependency on mpi4py (#8). As far as I know we currently do not have a use-case, where MPI is really required for the python bindings. We already made good experiences entirely removing the MPI from the python bindings (https://precice.discourse.group/t/installation-of-python-bindings-on-cray/149/4?u=benjaminrueth).

Therefore, I would suggest to entirely remove MPI from the python bindings. If anybody knows a use-case, where we cannot remove MPI from the python bindings as suggested in this PR, please feel free to comment.


uekerman:
> As far as I know we currently do not have a use-case, where MPI is really required for the python bindings.

Testing a simple parallel solver dummy should be fairly easy.

BenjaminRueth:
> > As far as I know we currently do not have a use-case, where MPI is really required for the python bindings.
> 
> Testing a simple parallel solver dummy should be fairly easy.

:+1: I will add a test for this.

solverdummy/README.md
```diff
@@ -10,6 +10,10 @@ You can test the dummy solver by coupling two instances with each other. Open tw
  * `python3 solverdummy.py precice-config.xml SolverOne MeshOne`
  * `python3 solverdummy.py precice-config.xml SolverTwo MeshTwo`
 
+## parallel version
+
+For the parallel verison of the solverdummy, you have to run add the command `mpirun -n <N>` to the  commands from above.
```
MakisH:
```suggestion
For the parallel verison of the solverdummy, you have to add the command `mpirun -n <N>` to the  commands from above.
```

BenjaminRueth:
> 
> 
> > > As far as I know we currently do not have a use-case, where MPI is really required for the python bindings.
> > 
> > 
> > Testing a simple parallel solver dummy should be fairly easy.
> 
> 👍 I will add a test for this.

Test works. Also without MPI.

setup.py
```diff
@@ -21,53 +21,13 @@
...
-    else:  # if PRECICE_MPI_IMPLEMENTATION is not mpich or openmpi quit.
-        raise Exception("unknown/no mpi found using compiler %s. Could not build PySolverInterface." % mpi_compiler_wrapper)
-
-    return mpi_compile_args, mpi_link_args
-
-
-def get_extensions(mpi_compiler_wrapper, is_test):
+def get_extensions(is_test):
     compile_args = []
     link_args = []
```
fsimonis: This list can now contain only a single argument. Maybe there is a more elegant way of writing this?

BenjaminRueth: The one-element-list is indeed a bit strange, but I currently do not really see a nicer solution. Note that we also want to be able to provide an empty list, if we are running the tests. I would keep it as it is.

Issue #37, opened 2020-02-10
Provide wheels via PyPI for faster installation


BenjaminRueth:
See, for example https://github.com/marketplace/actions/python-wheels-manylinux-build

Issue #15, opened 2020-01-14
Proper treatment of mpi4py?


BenjaminRueth:
Currently we throw a warning (see https://github.com/precice/python-bindings/commit/29dc71f6aa6f541c8b6bc88376b25965dfda3647), if `mpi4py` is imported before `precice`, since this may lead to errors.

## Example

See [this revision](https://github.com/precice/elastictube1d/tree/1689963b1299978be9630f86ae4722127e43da49) of the elastic tube, where executing `FluidSolver.py`leads to a Segmentation Fault due to [wrong order of imports](https://github.com/precice/elastictube1d/blob/1689963b1299978be9630f86ae4722127e43da49/python/FluidSolver.py#L16-L18):

```
from mpi4py import MPI
import precice
from precice import *
```

Changing this to

```
import precice
from precice import *
from mpi4py import MPI
```

eliminates the error.

## What do we want?

It would be better to be able to support both implementations from above, since this would be more user friendly.

BenjaminRueth:
Here is the error that we currently get (note the warning in the second line):

```
benjamin@benjamin-ThinkPad-X1-Yoga-2nd:~/elastictube1d/python$ python3 FluidSolver.py
/usr/lib/python3.6/importlib/_bootstrap.py:219: UserWarning: Please don't import mpi4py before importing precice, since this may lead to errors. If you need to import mpi4py, you can do so after importing precice.
  return f(*args, **kwds)
Starting Fluid Solver...
N: 100
Configure preCICE...
[2020-01-14 17:45:16.331860] [0x00007f6355836740] [trace]   Entering operator()
[2020-01-14 17:45:16.331894] [0x00007f6355836740] [trace]   Leaving initializeMPI
(0) 17:45:16 [impl::SolverInterfaceImpl]:101 in configure: This is preCICE version 1.6.1
(0) 17:45:16 [impl::SolverInterfaceImpl]:102 in configure: Revision info: v1.6.1-208-g625be7cb
(0) 17:45:16 [impl::SolverInterfaceImpl]:103 in configure: Configuring preCICE with configuration: "precice-config.xml"
[benjamin-ThinkPad-X1-Yoga-2nd:22012] *** Process received signal ***
[benjamin-ThinkPad-X1-Yoga-2nd:22012] Signal: Segmentation fault (11)
[benjamin-ThinkPad-X1-Yoga-2nd:22012] Signal code: Address not mapped (1)
[...]

Issue #12, opened 2019-12-10
Sort out dependencies of bindings


BenjaminRueth:
See https://github.com/precice/precice/pull/508

# Todo

- [x] can we avoid to [install numpy](https://github.com/precice/python-bindings/blob/7ddf2894644bb596e3ddbf772059ed98ab61b5ed/.travis.yml#L13) via apt? https://github.com/precice/python-bindings/commit/f51667763a585c0e736999a2f464e28aca2a8813
- [x] can we avoid to [install mpi](https://github.com/precice/python-bindings/blob/7ddf2894644bb596e3ddbf772059ed98ab61b5ed/.travis.yml#L11-L12) via apt? -> No

BenjaminRueth:
If we remove everything related to mpi from the python bindings, we do not need to install mpi via apt anymore (See https://github.com/precice/python-bindings/commit/7c8f67cd12698ff375cd060ed7eda804701f3cd5) . 

However, this might have some side effects which I am currently not aware of. #15 gives an idea of what might go wrong. 

Before just removing everything related to mpi from the bindings, we should also carefully evaluate the use-cases that we want to cover with the bindings. https://github.com/precice/precice/pull/146#pullrequestreview-148887126 gives some examples.

BenjaminRueth:
> If we remove everything related to mpi from the python bindings, we do not need to install mpi via apt anymore (See [7c8f67c](https://github.com/precice/python-bindings/commit/7c8f67cd12698ff375cd060ed7eda804701f3cd5)) .

I created an experimental draft for this approach on https://github.com/precice/python-bindings/tree/no_mpi.

BenjaminRueth:
As far as I understand, we cannot avoid installing MPI via apt. See https://github.com/precice/python-bindings/pull/41#issuecomment-584772729. Also refer to the documentation of [`mpi4py`](https://pypi.org/project/mpi4py/3.0.3/#description): One of the first conditions mentioned in the **Install** paragraph is that you need *a working MPI implementation and the mpicc compiler wrapper is on your search path* before running `pip install mpi4py`.

Closing this issue.

PR #41, opened 2020-02-11
Don't install MPI for tests

BenjaminRueth:


BenjaminRueth:
It seems like we cannot avoid installing `libopenmpi-dev` and/or `openmpi-bin`. Note that we already face issues, when we try to install the `mpi4py` dependency of the python bindings, since it cannot find `mpi.h` during its build process. This means a few different things:

* If we require `mpi4py` as a dependency, we implicitly also require an mpi installation. Strongly related to #15 
* I assume that `mpi4py` does not provide wheels, since it has to link against the system's mpi. This means that what is proposed in #37 might actually not work for us, since the situation with preCICE and the python-bindings is very similar to the situation with MPI and mpi4py.

BenjaminRueth:
We cannot avoid installing MPI via apt. See https://github.com/precice/python-bindings/issues/12.

PR #56, opened 2020-09-27
Support empty mesh for set_mesh_vertex/vertices

### LINKED ISSUE #55: Defining empty mesh in the interface gives an error
For parallel simulations, not every rank will consist a part of the coupling interface and hence the ranks which have no coupling interface will define meshes in preCICE using an empty list or empty numpy array. For example if `set_mesh_vertices()` is called with an empty list or empty array in place of the intended function argument for mesh coordinates, the following error is seen:
```
File "precice.pyx", line 519, in precice.Interface.set_mesh_vertices
ValueError: need more than 1 value to unpack
```
This scenario needs to be handled properly in the bindings. If it is intended that such empty ranks do not attempt to set up a mesh in preCICE then a proper error message would help here.

BenjaminRueth:
* Adds proper handling for empty list and tuple
* Adds corresponding tests
* Closes #55

### DIFF
--- a/precice.pyx
+++ b/precice.pyx
@@ -446,8 +446,12 @@ cdef class Interface:
         """
         if not isinstance(position, np.ndarray):
             position = np.asarray(position)
-        dimensions = position.size
-        assert(dimensions == self.get_dimensions())
+        if len(position) > 0:
+            dimensions = position.size
+            assert(dimensions == self.get_dimensions())
+        elif len(position) == 0:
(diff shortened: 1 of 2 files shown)


IshaanDesai (approved):
I checked this implementation against a [parallel prototype of the fenics-adapter](https://github.com/precice/fenics-adapter/tree/parallelMidLevel) and `set_mesh_vertices` call now works on ranks which have no part of the coupling boundary.

PR #63, opened 2020-10-28
Patching setup.py for installation without pip for Spack

ajaust:
For the installation of the bindings with Spack I have patched `setup.py`. The Spack package (https://github.com/spack/spack/pull/19558) uses the `setup.py` directly instead of using `pip`.

However, for checking the version I use a crude fix by importing `from setuptools._vendor.packaging`. I had problems when using the packaging package from Spack.


ajaust:
When I add `py-packaging` explicitly to the Spack-recipe, I end up with the following error

```
==> py-pyprecice: Executing phase: 'build_ext'
==> [2020-10-28-11:22:35.261548] '/home/jaustar/software/spack/spack-fork/opt/spack/linux-ubuntu18.04-skylake/gcc-7.5.0/python-3.8.6-c5r3eojydom26iptzff22kt3hva5gep4/bin/python3.8' '-s' 'setup.py' '--no-user-cfg' 'build_ext' '--include-dirs=/home/jaustar/software/spack/spack-fork/opt/spack/linux-ubuntu18.04-skylake/gcc-7.5.0/precice-2.0.2-zxbytdw2k5ofq6ju5k434iv4zidfj4vd/include' '--library-dirs=/home/jaustar/software/spack/spack-fork/opt/spack/linux-ubuntu18.04-skylake/gcc-7.5.0/precice-2.0.2-zxbytdw2k5ofq6ju5k434iv4zidfj4vd/lib'
Traceback (most recent call last):
  File "setup.py", line 4, in <module>
    from packaging import version
ModuleNotFoundError: No module named 'packaging'
```

setup.py
```diff
@@ -1,15 +1,20 @@
 import os
 import subprocess
 import warnings
-from packaging import version
-import pip
+from setuptools._vendor.packaging import version
```
BenjaminRueth, line 4: Like already discussed: I do not really like using `setuptools._vendor.packaging` and would prefer `packaging`. https://github.com/precice/python-bindings/pull/63#issuecomment-717839037 is, of course, an argument to patch this on the spack-side, if necessary. I'm a bit surprised that this does not work. Any idea?

setup.py
```diff
@@ -1,15 +1,20 @@
...
+try:
+    import pip
+    if version.parse(pip.__version__) < version.parse("19.0"):
+        # version 19.0 is required, since we are using pyproject.toml for definition of build-time depdendencies. See https://pip.pypa.io/en/stable/news/#id209
+        warnings.warn("You are using pip version {}. However, pip version > 19.0 is recommended. You can continue with the installation, but installation problems can occour. Please refer to https://github.com/precice/python-bindings#build-time-dependencies-cython-numpy-defined-in-pyprojecttoml-are-not-installed-automatically for help.".format(pip.__version__))
+
+    if version.parse(pip.__version__) < version.parse("10.0.1"):
+        warnings.warn("You are using pip version {}. However, pip version > 10.0.1 is required. If you continue with installation it is likely that you will face an error. See https://github.com/precice/python-bindings#version-of-pip3-is-too-old".format(pip.__version__))
+except:
+    warnings.warn("Assuming that you are not using pip!")
```
BenjaminRueth, line 16: I think this is a valid use-case (also outside of the "spack world"), since a user might just run `python3 setup.py install`. Then `pip` is not required. As far as I know running `python3 setup.py install` bypasses the dependencies defined in `pyproject.toml`. I would not fix this by duplicating the content of `pyproject.toml`, since the recommended approach is to use `pip`.

We should instead throw a warning, if `pip` is not used:

```suggestion
try:
    import pip
    if version.parse(pip.__version__) < version.parse("19.0"):
        # version 19.0 is required, since we are using pyproject.toml for definition of build-time depdendencies. See https://pip.pypa.io/en/stable/news/#id209
        warnings.warn("You are using pip version {}. However, pip version > 19.0 is recommended. You can continue with the installation, but installation problems can occour. Please refer to https://github.com/precice/python-bindings#build-time-dependencies-cython-numpy-defined-in-pyprojecttoml-are-not-installed-automatically for help.".format(pip.__version__))

    if version.parse(pip.__version__) < version.parse("10.0.1"):
        warnings.warn("You are using pip version {}. However, pip version > 10.0.1 is required. If you continue with installation it is likely that you will face an error. See https://github.com/precice/python-bindings#version-of-pip3-is-too-old".format(pip.__version__))
except:
    warnings.warn("Assuming that you are not using pip! If the bindings work correctly, you do not have to do anything. If you face problems during installation or running pyprecice, this means that you have to make sure that all dependencies are installed correctly and repeat the installation of pyprecice. Refer to pyproject.toml for a list of dependencies.")
```

BenjaminRueth (changes requested):

setup.py
ajaust: I would assume that this is either a Spack issue or a user (my) issue using Spack. If I use the branch where the `packaging` module is installed, I can access `packaging` in an interactive python session, but the Spack recipe does not find it. When I set up the recipe I simply gave up at some point and went for the crude fix. (I would prefer a proper fix though.)

BenjaminRueth: I installed `py-packaging` manually in my spack env and observed the same issue using [your branch](https://github.com/ajaust/spack/tree/version-from-packaging):
```
==> Installing py-pyprecice
==> No binary for py-pyprecice found: installing from source
==> Using cached archive: /home/benjamin/spack/var/spack/cache/_source-cache/archive/97/972f574549344b6155a8dd415b6d82512e00fa154ca25ae7e36b68d4d2ed2cf4.tar.gz
==> py-pyprecice: Executing phase: 'build_ext'
==> Error: ProcessError: Command exited with status 1:
    '/home/benjamin/spack/opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/python-3.8.6-wlcjv5qbqqiqwfkkkksxjyh2k2ru2hvw/bin/python3.8' '-s' 'setup.py' '--no-user-cfg' 'build_ext' '--include-dirs=/home/benjamin/spack/opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/precice-2.1.1-utdkrqewi3yphbgst2itc7ak4l2puu2w/include' '--library-dirs=/home/benjamin/spack/opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/precice-2.1.1-utdkrqewi3yphbgst2itc7ak4l2puu2w/lib'
See build log for details:
  /tmp/benjamin/spack-stage/spack-stage-py-pyprecice-2.1.1.1-ummpnvtonm5vajataudie3yx2vu3vnjc/spack-build-out.txt

benjamin@benjamin-ThinkPad-X1-Yoga-2nd:~/spack$ cat /tmp/benjamin/spack-stage/spack-stage-py-pyprecice-2.1.1.1-ummpnvtonm5vajataudie3yx2vu3vnjc/spack-build-out.txt 
==> py-pyprecice: Executing phase: 'build_ext'
==> [2020-10-28-12:11:27.387199] '/home/benjamin/spack/opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/python-3.8.6-wlcjv5qbqqiqwfkkkksxjyh2k2ru2hvw/bin/python3.8' '-s' 'setup.py' '--no-user-cfg' 'build_ext' '--include-dirs=/home/benjamin/spack/opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/precice-2.1.1-utdkrqewi3yphbgst2itc7ak4l2puu2w/include' '--library-dirs=/home/benjamin/spack/opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/precice-2.1.1-utdkrqewi3yphbgst2itc7ak4l2puu2w/lib'
Traceback (most recent call last):
  File "setup.py", line 4, in <module>
    from packaging import version
ModuleNotFoundError: No module named 'packaging'
benjamin@benjamin-ThinkPad-X1-Yoga-2nd:~/spack$ python3 -c "import packaging; print(packaging)"
<module 'packaging' from '/home/benjamin/spack/var/spack/environments/myproject/.spack-env/view/lib/python3.8/site-packages/packaging-19.2-py3.8.egg/packaging/__init__.py'>
```

ajaust: If you use my branch, it should install `py-packaging` anyway since I specify it as a dependency for `py-pyprecice`.

I added a remark in the Spack PR. Maybe they know more.

BenjaminRueth, line 4: I would suggest to rollback this change and (if necessary) patch on the spack side. Ideally we can somehow fix the issue mentioned above in the spack package.

```suggestion
from packaging import version
```

BenjaminRueth (changes requested):
To summarize: With the suggested changes I think we are close to merging this PR. If necessary, we can open another one, if the `packaging` issue cannot be resolved on the spack side.

setup.py
BenjaminRueth: See https://github.com/precice/python-bindings/pull/63#discussion_r513398057. I think we should try to solve this issue either on the spack-side or in another PR. Ok? Then we can "resolve" this conversation.

ajaust: The change has been rolled back in 5b35c3cc02677ed198d57219415bdd639d9d1907

setup.py
ajaust: This has been incorporated in b72f5cc186f5889702232315905fda79d14e3a5a

ajaust:
I started to wonder whether the pip version should be checked in the `setup.py` at all. The minimum required version of pip should probably stated clearly in the installation instructions and that's it. There does not seem to exist any standard way to check whether the installation is triggered by `pip` or any other tool as far as I know. This leads me to the conclusion that there should not be any dependency on a tool like `pip` in the `setup.py`.

setup.py
BenjaminRueth: Fixed in 5b35c3c.

BenjaminRueth:
What was the current status of this PR? As far as I remember it's ready to merge. @ajaust Or are there any open points? If not, I will approve and merge.

ajaust:
I am not sure. How small should the fixes be? 

The PR relaxes the dependency of `pip`. This might be done in a non-standard/unstable manner via `if "pip" in __file__:` so I am not sure if I am a fan of that. This might give us trouble some later time. However, I have no idea how Python packages solve this problem normally. At the same time there is work going on in https://github.com/precice/python-bindings/issues/64 so I lost a bit track what part of `setup.py` should be fixed where and when.  

Regarding the Spack installation this is only a small step for full Spack support. At the moment we still need the patch and a adapted installation procedure for the Spack package since:

- `from packaging import version` does not work and needs to be removed when using Spack (thus a patch is needed anyway).
- The current `setup.py` is lacking some options Spack expects such that we need to overwrite the Spack installation routine of the `py-pyprecice` package (https://github.com/precice/python-bindings/issues/64). 

As the Spack package needs a patch anyway, we could also change the patch in the Spack repository to remove the whole `pip` related procedure. I am unsure what would be the best approach

1. Fixing this small part here and adapting the patch in the Spack repository accordingly.
2. Leaving the `setup.py` as it is at the moment and fix it properly such that we do not need the patch in the Spack repository anymore and such that we can use the standard installation procedure of Spack.

BenjaminRueth:
Thanks for the summary!

> The PR relaxes the dependency of `pip`. This might be done in a non-standard/unstable manner via `if "pip" in __file__:` so I am not sure if I am a fan of that. This might give us trouble some later time. However, I have no idea how Python packages solve this problem normally. 

Relaxing the dependency on `pip` is a good thing, since this will allow us reduce the code that needs a [patch](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/py-pyprecice/deactivate-version-check-via-pip.patch). The non-standard `if "pip" in __file__:` is a bit a downside, but I would suggest to accept it, since we currently do not see a better solution. 

Comparing the two options that we currently have (dependency on `pip` vs. non-standard check for `pip`), I think the latter is preferable and therefore would like to merge this PR. Ok?

> At the same time there is work going on in #64 so I lost a bit track what part of `setup.py` should be fixed where and when.
> 
> Regarding the Spack installation this is only a small step for full Spack support. At the moment we still need the patch and a adapted installation procedure for the Spack package since:
> 
>   * `from packaging import version` does not work and needs to be removed when using Spack (thus a patch is needed anyway).

I think that here [the patch](https://github.com/spack/spack/blob/f041876470ec9f9a1c38b05f080785955582098f/var/spack/repos/builtin/packages/py-pyprecice/deactivate-version-check-via-pip.patch#L9-L11
) that we provide is a good solution: Using `packaging` is the best solution that unfortunately does not work on spack. So patching to use `from setuptools._vendor.packaging import version` it is a good workaround and the only option we have.

At the current state we do not have to do anything / cannot do anything here.

>   * The current `setup.py` is lacking some options Spack expects such that we need to overwrite the Spack installation routine of the `py-pyprecice` package (#64).

Here, #65 will hopefully converge at some point. Right now, we are a bit stuck in this PR, but might occasionally find a solution.

> As the Spack package needs a patch anyway, we could also change the patch in the Spack repository to remove the whole `pip` related procedure. I am unsure what would be the best approach
> 
>   1. Fixing this small part here and adapting the patch in the Spack repository accordingly.
> 
>   2. Leaving the `setup.py` as it is at the moment and fix it properly such that we do not need the patch in the Spack repository anymore and such that we can use the standard installation procedure of Spack.

I assume that we will still need a patch (at least due to `packaging`). #65 is also still open. Let's hope that we will be able to solve this, but right now I cannot really estimate this.

My suggestion: Let's merge this PR and create a draft PR to update [the patch](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/py-pyprecice/deactivate-version-check-via-pip.patch) accordingly.

BenjaminRueth (commented):
I did a few checks with older pip version (`pip 9.0..2`) and using `python3 setup.py install`. All warnings are issued correctly. However, there is a downside: `pip` seems to only show warnings, if the `-v` option is provided.

Maybe we are really following a wrong approach here? If the user does not see the warning in the default case, it is not worth a lot to provide a warning.

BenjaminRueth:
The latest commit introduces a fundamental change: pip 19.0.0 is now the minimal requirement. This allows us to raise an exception, if the version is too low and this is also something the user will see.

BenjaminRueth (approved):
From my point of view the new, stricter requirements on the pip version are reasonable and guarantee safe usage.

BenjaminRueth:
@fsimonis can you take a look at this PR? Especially w.r.t code sanity and documentation of the installation process, requirements and dependencies?

setup.py
```diff
@@ -1,14 +1,35 @@
 import os
 import warnings
 from packaging import version
```
fsimonis, line 3: This will still fail under some circumstances. Correct?

Why not simply `try:` to `import packaging` and skip the checks when necessary.

BenjaminRueth: After merging #70 we will only need `version` for comparing versions w.r.t some warnings. I think we can easily use `try` here and if it fails issue a warning `some warnings are not shown, since the module packaging could not be found on your system`.

I can take care of implementing this here.

BenjaminRueth: done in 73451b8.

BenjaminRueth:
I merged the current state of develop into this fork. From my perspective everything now looks clean. I would like to merge this PR soon and then continue with #70.

BenjaminRueth:
Already merged #70, since this will allow us to relax on the `packaging` dependency here. I will merge `develop` into this PR now and then merge #63.

PR #66, opened 2020-10-30
add Spack to documentation

BenjaminRueth:
https://github.com/spack/spack/pull/19558

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## latest
 
-* Bindings are now available via spack. https://github.com/spack/spack/pull/19558
+* Bindings are now available via Spack. https://github.com/spack/spack/pull/19558
 
 ## 2.1.1.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
+## latest
+
+* Bindings are now available via spack. https://github.com/spack/spack/pull/19558
+
 ## 2.1.1.1
 
 * Bindings can now handle mesh initialization with no vertices. This behavior is consistent with the C++ preCICE API.
--- a/README.md
+++ b/README.md
@@ -56,9 +56,9 @@ If preCICE (the C++ library) was installed in a custom prefix, or only built but
 
 The precice wiki provides more informaiton on [linking preCICE](https://github.com/precice/precice/wiki/Linking-to-preCICE).
 
-## Using spack
+## Using Spack
 
-Use `spack install py-pyprecice` to install the spack package.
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -56,6 +56,10 @@ If preCICE (the C++ library) was installed in a custom prefix, or only built but
 
 The precice wiki provides more informaiton on [linking preCICE](https://github.com/precice/precice/wiki/Linking-to-preCICE).
 
+## Using spack
+
+Use `spack install py-pyprecice` to install the spack package.
```
MakisH, line 61: I don't really want to try this on my system, as I am afraid it would take a while to build. But maybe a comment for my case, as well as a link to our documentation page explaning what Spack is etc would help.

So, potentially helpful additions:
- When should one need this? (a: when getting preCICE from Spack, or for people not having a recent enough Python version)

Follow-ups would be (but probably they are summarized in "only use this if you installed preCICE from Spack":
- What to do when you already have preCICE installed from another source? (a: then probably use `spack install py-pyprecice~precice` or consider using pip)
- What to do when you already have Python installed?

Maybe just adding a note that "this will also install preCICE and Python from Spack" would be enough of a warning.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
+## latest
+
+* Bindings are now available via spack. https://github.com/spack/spack/pull/19558
```
MakisH, line 7:
```suggestion
* Bindings are now available via Spack. https://github.com/spack/spack/pull/19558
```

MakisH (approved):
Nice to have this addition! I think a comment for the use case, what this actually does, and linking to our general page for Spack would be helpful.

Feel free to merge after these changes.

README.md
ajaust: Maybe we can add a link to https://spack.readthedocs.io/en/latest/getting_started.html#system-packages or mention that one can integrate available (system) packages in Spack. Not sure whether it would fit in our documentation or in the documentation of preCICE. Maybe that is an option to minimize compilation times for @MakisH as well. ;)

MakisH: We do have this page: https://github.com/precice/precice/wiki/preCICE-with-Spack

ajaust: In order to contribute to this point here. I would just say that a Spack-package is available named `py-pyprecice` without giving the installation command and then pointing the users to the Spack homepage or the preCICE documentation about Spack.  

I would expect that users who know Spack will be happy to see the package name and know how to proceed. Users who are new to Spack have to get used to Spack themselves. A sole `spack install py-precice` might not be the right command for them to install the package. Maybe they have to configure stuff beforehand or they want to use environments etc. 

This way we also avoid duplicating information from the Spack documentation. Spack commands may change which would give us extra work updating our documentation.

ajaust (changes requested):
I agree with @MakisH that we should do something about the installation procedure using Spack.

README.md
BenjaminRueth: Thanks for the feedback! I hope that 026926f somehow goes into the right direction. Asked @ajaust for another review.

ajaust (approved):
Looks good. :-)

PR #65, opened 2020-10-30
Clean-up of setup.py

BenjaminRueth:
closes #64

PR #67, opened 2020-11-11
Add tests to github actions CI

### LINKED ISSUE #60: Use Github Actions CI
preCICE CI was moved to Github Actions in https://github.com/precice/precice/pull/843. We should also move our testing infrastructure for the the python bindings to Github Actions.

BenjaminRodenberg:
Closes #60 

This PR adds tests for different building and installation options. This is especially useful for making sure that we do not break requirements for the spack image (such as the option `--single-version-externally-managed`).

# Failing tests:

- [x] `setup_install` -> Will be fixed by updating dependencies on `develop`
- [x] `setup_install_single_version_externally_managed` -> Will be fixed via #65


BenjaminRueth:
@fsimonis does the PR look more or less sane w.r.t usage of github actions? I will also open a PR in https://github.com/precice/ci-images in a moment.

fsimonis:
Wouldn't it be a bit more resource-friendly to run all the installation versions only if the solverdummy test succeeds?

Some additional notes:
* You can give your jobs real names without underscores.
* You may have to use a custom shell when running the actions in the docker container. https://github.com/precice/precice/blob/641e5057cb5d71d39ed43ae1dbc6af597b60bf52/.github/workflows/build-and-test.yml#L20-L22

BenjaminRueth:
Tests finally working. Ready for review.

@fsimonis: Can you quickly check whether everything is "sane" from a github-actions-point-of-view?
@ajaust: Are we missing something important that we need for spack?

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,115 @@
...
+    name: Run setup build_ext
+    runs-on: ubuntu-latest
+    container: benjaminrueth/precice
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - name: Install & upgrade pip3
+        run: |
+          apt-get install -y python3-pip
+          pip3 install --upgrade --user pip
```
ajaust, line 15: Is it considered safe to mix the `pip` version distributed via the Ubuntu repository with a user-installed and updated version? I remember having troubles with that some time in the past (or a similar setup).

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,115 @@
...
+    name: Run setup install
+    runs-on: ubuntu-latest
+    container: benjaminrueth/precice
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - name: Install & upgrade pip3
+        run: |
+          apt-get install -y python3-pip
+          pip3 install --upgrade --user pip
```
ajaust, line 33: Is it considered safe to mix the `pip` version distributed via the Ubuntu repository with a user-installed and updated version? I remember having troubles with that some time in the past (or a similar setup).

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,115 @@
...
+    name: Run setup install --single-version-externally-managed (for spack)
+    runs-on: ubuntu-latest
+    container: benjaminrueth/precice
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - name: Install & upgrade pip3
+        run: |
+          apt-get install -y python3-pip
+          pip3 install --upgrade --user pip
```
ajaust, line 53: Is it considered safe to mix the `pip` version distributed via the Ubuntu repository with a user-installed and updated version? I remember having troubles with that some time in the past (or a similar setup).

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,115 @@
...
+    name: Run setup test
+    runs-on: ubuntu-latest
+    container: precice/ci-ubuntu-1804
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - name: Install & upgrade pip3
+        run: |
+          apt-get update && apt-get install -y python3-pip
+          pip3 install --upgrade --user pip
```
ajaust, line 71: Why do you `apt-get update` here before installing `pip`, but not for the other configurations?

And also:
Is it considered safe to mix the `pip` version distributed via the Ubuntu repository with a user-installed and updated version? I remember having troubles with that some time in the past (or a similar setup).

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,115 @@
...
+    name: Run pip install
+    runs-on: ubuntu-latest
+    container: benjaminrueth/precice
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - name: Install & upgrade pip3
+        run: |
+          apt-get install -y python3-pip
+          pip3 install --upgrade --user pip
```
ajaust, line 94: Is it considered safe to mix the `pip` version distributed via the Ubuntu repository with a user-installed and updated version? I remember having troubles with that some time in the past (or a similar setup).

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,115 @@
...
+    name: Run solverdummy
+    runs-on: ubuntu-latest
+    container: benjaminrueth/precice
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - name: Install & upgrade pip3
+        run: |
+          apt-get install -y python3-pip
+          pip3 install --upgrade --user pip
```
ajaust, line 110: Is it considered safe to mix the `pip` version distributed via the Ubuntu repository with a user-installed and updated version? I remember having troubles with that some time in the past (or a similar setup).

ajaust (commented):
I have added some comments. It is mainly about updating `pip` when it is installed via the system's package manager.

.github/workflows/build-and-test.yml
BenjaminRodenberg: As far as I know everything is fine as long as the `--user` flag is used. One thing that caused me a lot of trouble was `sudo pip3 ...`, since this changes the system pip.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,119 @@
+name: Build and Test
+on: push
+      
+jobs:  
+  setup_for_spack:
+    name: Run setup.py phases needed by spack
+    runs-on: ubuntu-latest
+    container: benjaminrueth/precice
```
fsimonis, line 8: Shouldn't the container be `container: precice/precice`?

fsimonis (approved):
This PR looks good to me.
I suggest to merge it and see how it behaves on develop.
The action can always be tweaked if necessary.

One last idea: would it make sense to define dependencies between the tests to limit the amount of failing ones?

.github/workflows/build-and-test.yml
BenjaminRodenberg: I tried to remove the `apt-get update` and it did not work. Note that we use `precice/ci-ubuntu-1804` for this test and `precice/precice` for the other tests. I think that this is the reason. [`precice/ci-ubuntu-1804` runs `apt-get update` and then deletes the cache](https://github.com/precice/systemtests/blob/d39f4d89d69f18c9de9f8bd1f08633a68bcda9a2/precice/Dockerfile.Ubuntu1604.home#L6-L17) while [`precice/precice` does not delete the cache](https://github.com/precice/precice/blob/69edf625fe202e1006e4c91431ba923290e22e01/tools/releasing/packaging/docker/Dockerfile#L6-L7).

BenjaminRodenberg: Call `apt-get update` for all test now. 958aa3f

I also opened https://github.com/precice/precice/pull/927 resulting from this discussion. Thanks for the useful remark @ajaust!

.github/workflows/build-and-test.yml
BenjaminRodenberg: Fixed in c824ac1

BenjaminRodenberg:
> This PR looks good to me.
> I suggest to merge it and see how it behaves on develop.
> The action can always be tweaked if necessary.
> 
> One last idea: would it make sense to define dependencies between the tests to limit the amount of failing ones?

Yes and no. I like it that all tests are run in parallel and therefore finish after approx. 5 min (longest runtime: `build_spack`). A long sequential chain of tests will have a longer runtime. I think having "batches" of 2 depending jobs will be a good compromise (first the general one, then the more specific. First `setup_install`, then `setup_install_single_version_externally_managed`). I will implement this and then merge.

Issue #50, opened 2020-07-20
Support __version__


BenjaminRueth:
Currently, we cannot get the version of the python bindings via the following code:

```
import precice
print precice.__version__
```

Thanks to @ajaust for pointing this out!

PR #68, opened 2020-11-13
Test spack build with CI and provide instructions and tools for release

BenjaminRueth:
This PR aims at providing a CI pipeline for testing the spack build process using the current version of the python bindings and a `package.py` that is provided in this repository.

# Open Todo

- [x] Provide a docker image with the necessary dependencies to reduce build time


BenjaminRueth:
https://github.com/precice/python-bindings/pull/65#issuecomment-727237472 provides an update for `package.py` and another `patch` that we will have to implement here as soon as #65 is merged.

BenjaminRueth (commented):
Quick summary of what this PR actually does (at the current state):

### General Problem

We do not test whether the python-bindings can be built using spack. The corresponding package `py-pyprecice` is only updated for new releases. This PR adds a CI workflow in github actions for building the spack package based on a given state of the repository (`develop` or any other branch).

### Details on `build-spack.yml`

We use the image `benjaminrueth/ci-spack-precice-1804` as a basis to reduce build time. This image already provides spack and the dependencies of preCICE:

```
# Dockerfile of benjaminrueth/ci-spack-precice-1804

FROM spack/ubuntu-bionic:latest
RUN spack install boost@1.74.0 cmake@3.18.4 eigen@3.3.8 petsc@3.14.1
RUN spack env create precice
```

### What in the test?

We provide all necessary files for the spack package under `spack/var/spack/repos/builtin/packages/py-pyprecice`. The provided `package.py` and `*.patch` files are used to build the spack package based on the current state of the repository. The script `jinja-instantiate.py` allows to specify the branch to be used for `py-pyprecice@develop`.

We finally copy these files to the spack directory of `benjaminrueth/ci-spack-precice-1804` to overwrite the specification from `spack/develop` and install the package via `spack install py-pyprecice@develop`.

### Current issues

Unfortunately github actions do not pick up the dependencies provided within `benjaminrueth/ci-spack-precice-1804`, while using `act` or when running the container interactively everything looks fine. This leads to a huge build time.

fsimonis:
I had the same issue with the precice actions. You need to use a login shell:

https://github.com/precice/precice/blob/641e5057cb5d71d39ed43ae1dbc6af597b60bf52/.github/workflows/build-and-test.yml#L20-L22

BenjaminRueth:
### Notes for the review

@ajaust: any obvious spack flaws?
@fsimonis: what do you think about the general workflow and test? We can also consider moving some of the docker/CI related things to https://github.com/precice/ci-images, if you think this makes sense. Building the image with the dependencies, of course, takes ages, but this is something one only has to do from time to time.

spack/ci-spack-pyprecice-deps-1804.dockerfile
```diff
@@ -0,0 +1,10 @@
+# Build stage with Spack pre-installed and ready to be used
+FROM spack/ubuntu-bionic:latest
+RUN spack env create precice
+RUN spack env activate precice
+RUN spack install boost@1.74.0 cmake@3.18.4 eigen@3.3.8 petsc@3.14.1
+RUN spack install pkgconfig
+RUN spack env activate precice && spack concretize -f && spack install pkgconfig && spack install precice@develop
+
+# Install python dependencies for pyprecice
+RUN spack install py-setuptools py-wheel py-cython py-numpy py-mpi4py
```
ajaust, line 10: Why are you installing pkgconfig in line 6 and 7?

I think the cleanest way of compiling/adding the dependencies to the environment and concretizing them trying to avoid duplication would be:
```
RUN spack env activate precice \ 
    && spack add boost@1.74.0 cmake@3.18.4 eigen@3.3.8 petsc@3.14.1 pkgconfig precice@develop py-setuptools py-wheel py-cython py-numpy py-mpi4py \
    && spack concretize -f \
    && spack install
```

You might specify the environment in a `spack.yaml` file in `${SPACK_ROOT}/var/spack/environments/precice` (where `precice` is the name of the environment). It might help to add the line `concretization: together` in case dependencies are duplicated.

The `spack.yaml` might look like this:
```
# This is a Spack Environment file.
#
# It describes a set of packages to be installed, along with
# configuration settings.
spack:
   specs:
  - boost@1.74.0 
  - cmake@3.18.4 
  - eigen@3.3.8 
  - petsc@3.14.1
  - pkgconfig
  - precice@develop
  - py-setuptools 
  - py-wheel 
  - py-cython 
  - py-numpy
  - py-mpi4py
  view: true
  concretization: together
```

If it works for now, I guess it should be fine. I am not sure if @fsimonis  has an opinion on how to set environments up properly.

spack/var/spack/repos/builtin/packages/py-pyprecice/package.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def build_ext_args(self, spec, prefix):
+        return [
+            "--include-dirs=" + spec["precice"].headers.directories[0],
+            "--library-dirs=" + spec["precice"].libs.directories[0]
+        ]
+
+    #if self.spec.satisfies("@:2.1.1.1"):  # syntax is wrong
+    #    def install(self, spec, prefix):
+    #        self.setup_py("install", "--prefix={0}".format(prefix))
```
ajaust, line 64: Don't we need this for older versions of the bindings? 

If yes, I guess we should try to fix it properly. If we do not need that for older versions, it should be removed.

ajaust (commented):
I have added some comments, but it looks quite good to me.

spack/README.md
```diff
@@ -0,0 +1,19 @@
+# Spack package: py-pyprecice
+
+The spack package `py-pyprecice` provides the python bindings via spack and was submitted via https://github.com/spack/spack/pull/19558. This folder contains the spack package script that can be used for testing whether the spack installation works.
```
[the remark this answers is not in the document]
BenjaminRueth: Done 4299c8c.

spack/README.md
```diff
@@ -0,0 +1,19 @@
...
+```
+
+## Docker image ci-spack-pyprecice-deps-1804
+
+The test `build_spack` in `.github/workflows/build-spack.yml` uses the image `benjaminrueth/ci-spack-pyprecice-deps-1804` to reduce build time. Use the file `ci-spack-pyprecice-deps-1804.dockerfile` from this folder for building this image:
+```
+docker build -f ci-spack-pyprecice-deps-1804.dockerfile -t USERNAME/ci-spack-pyprecice-deps-1804.dockerfile .
+docker push USERNAME/ci-spack-pyprecice-deps-1804.dockerfile
+```
+## When a new spack release is necessary
```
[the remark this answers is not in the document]
BenjaminRueth: Done 4299c8c.

spack/ci-spack-pyprecice-deps-1804.dockerfile
BenjaminRueth: I changed the dockerfile in 2e22d8c. I will rebuild and test the new file in a moment. But let's keep this threat open until @fsimonis has finished the review. I'm not very experienced with spack and therefore another opinion on the question on "how do I setup an environment?" probably does not harm.

BenjaminRueth: Note on `pkgconfig`: Currently preCICE needs `pkgconfig` for finding PETSc. As far as I understand the issue this dependency is currently missing in the package definition of `precice`. Therefore, I'm installing it manually.

spack/var/spack/repos/builtin/packages/py-pyprecice/package.py
BenjaminRueth: Totally forgot about this. 100% true. I will try to find out how to implement this in `package.py`. I think we will need it for supporting older version.

spack/ci-spack-pyprecice-deps-1804.dockerfile
ajaust: I guess I missed that this is Docker and every `RUN` is executed in a new container (or so...). So I assume the sole
```
RUN spack install pkgconfig 
```
will install the package in the root environment and 
```
spack env activate precice && spack concretize -f && spack install pkgconfig && spack install precice@develop
```
adds it to the `precice` environment. Regarding the resulting software this should not make any difference. Software that is installed in an environment ends up in the root environment as well.

spack/var/spack/repos/builtin/packages/py-pyprecice/package.py
BenjaminRueth: d981f41 implements support for older versions. However, testing this is not so easy right now, since I cannot easily check for the version (#50).

spack/var/spack/repos/builtin/packages/py-pyprecice/remove-packaging-and-pip.patch
```diff
@@ -0,0 +1,21 @@
...
+-    # version 19.0 is required, since we are using pyproject.toml for definition of build-time depdendencies. See https://pip.pypa.io/en/stable/news/#id209
+-    warnings.warn("You are using pip version {}. However, pip version > 19.0 is recommended. You can continue with the installation, but installation problems can occour. Please refer to https://github.com/precice/python-bindings#build-time-dependencies-cython-numpy-defined-in-pyprojecttoml-are-not-installed-automatically for help.".format(pip.__version__))
+-
+-if version.parse(pip.__version__) < version.parse("10.0.1"):        
+-    warnings.warn("You are using pip version {}. However, pip version > 10.0.1 is required. If you continue with installation it is likely that you will face an error. See https://github.com/precice/python-bindings#version-of-pip3-is-too-old".format(pip.__version__))
+-
++from setuptools._vendor.packaging import version
+ from setuptools import setup
+ from setuptools.command.test import test
+ from Cython.Distutils.extension import Extension
```
BenjaminRueth, line 21: We will have to review this patch, when #63 and #70 are merged.

BenjaminRueth: Updated the patch according to #70 in 4e88ef6.

BenjaminRueth: Removed the patch, since it is not needed anymore. 50a0196

spack/var/spack/repos/builtin/packages/py-pyprecice/package.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def build_ext_args(self, spec, prefix):
+        return [
+            "--include-dirs=" + spec["precice"].headers.directories[0],
+            "--library-dirs=" + spec["precice"].libs.directories[0]
+        ]
+
+    def install(self, spec, prefix):
+        if self.version <= Version("2.1.1.1"):
+            self.setup_py("install", "--prefix={0}".format(prefix))
```
ajaust, line 61: Does this work nownewer versions of the adapter or does it skip the installation step completely then?

spack/ci-spack-pyprecice-deps-1804.dockerfile
```diff
@@ -0,0 +1,8 @@
+# Build stage with Spack pre-installed and ready to be used
+FROM spack/ubuntu-bionic:latest
+RUN spack env create precice
+# Install dependencies for precice and py-pyprecice
+RUN spack env activate precice \ 
+    && spack add boost@1.74.0 cmake@3.18.4 eigen@3.3.8 petsc@3.14.1 pkgconfig precice@develop py-setuptools py-wheel py-cython py-numpy py-mpi4py \
+    && spack concretize -f \
+    && spack install
```
ajaust, line 8: Is this acceptable with regards to compile time or do we need to create a Docker image that already contains all the dependencies?

ajaust (commented):
I have added my feedback, but that is only minor things.

spack/ci-spack-pyprecice-deps-1804.dockerfile
BenjaminRueth: The image is compiled locally (once) an provided on dockerhub (https://hub.docker.com/r/benjaminrueth/ci-spack-pyprecice-deps-1804). This Dockerfile is only provided in case we want to rebuild the image.

spack/var/spack/repos/builtin/packages/py-pyprecice/package.py
BenjaminRueth: The current test runs on `develop > 2.1.1.1` and works. I ran an additional tests and to me it looks like `install` does what it is supposed to do:

Running `python3 -c "import precice; print(precice.__version__)"` not from project root, but from an independent folder works. So the package is found correctly (through the python installation), which means that the installation was executed correctly.

If I run `python3 -c "import precice; print(precice)"`, I also get a reasonable path: `<module 'precice' from '/opt/spack/var/spack/environments/precice/.spack-env/view/lib/python3.8/site-packages/precice/__init__.py'>`

I think this looks good from my perspective. We should still explicitly ask for a double check on this part, when we submit the new `package.py` to Spack, since our reviewer will probably know the answer. I myself are not 100% sure.

ajaust: That sounds good!

spack/ci-spack-pyprecice-deps-1804.dockerfile
ajaust: That sounds good, too. I was not aware of the actual workflow.

fsimonis (approved):

PR #70, opened 2020-11-19
Use python-versioneer to support __version__

### LINKED ISSUE #50: Support __version__

BenjaminRueth:
This PR restructures the structure of this repository in the following way:

* `cyprecice` provides the cython extension
* `test` provides (as before) the test
* `precice` provides a pure python package

This restructuring was necessary, since I could not find a nice way to implement support for `__version__` in a cython-based package. However, for a python based package (which `precice` now is) there is https://github.com/python-versioneer/python-versioneer, which comes with some additional features.

Closes #50


pyproject.toml
```diff
@@ -1,3 +1,3 @@
 [build-system]
 # PEP 518 - minimum build system requirements
-requires = ["setuptools", "wheel", "Cython>=0.29", "packaging", "pip", "numpy", "mpi4py"]
+requires = ["setuptools", "wheel", "Cython>=0.29", "packaging", "pip", "numpy", "mpi4py", "versioneer"]
```
ajaust, line 3: We have to make sure to add the new dependency in our Spack package. 

Is it reasonable to add a new dependency just for the package version?

cyprecice/cyprecice.pyx
```diff
@@ -5,11 +5,8 @@
 The python module precice offers python language bindings to the C++ coupling library precice. Please refer to precice.org for further information.
 """
 
+cimport cyprecice
```
ajaust, line 8: Does this mean we have to rename the package from `pyprecice` to `cyprecice` or is this just some internal naming?

precice/_version.py
```diff
@@ -0,0 +1,525 @@
...
+# This file helps to compute a version number in source trees obtained from
+# git-archive tarball (such as those provided by githubs download-from-tag
+# feature). Distribution tarballs (built by setup.py sdist) and build
+# directories (produced by setup.py build) will contain a much shorter file
+# that just contains the computed version number.
+
+# This file is released into the public domain. Generated by
+# versioneer-0.19 (https://github.com/python-versioneer/python-versioneer)
+
+"""Git implementation of _version.py."""
```
ajaust, line 11: Does this file give us Git as a dependency for the package? Will it work if you use the tarball instead of the `python-bindings` Git repository?

setup.py
```diff
@@ -1,6 +1,7 @@
 import os
 import warnings
 from packaging import version
+import versioneer
```
ajaust, line 4: Is there any overlap in the functionality of these two packages? Can we achieve the same by only using one of the two?

setup.py
```diff
@@ -1,6 +1,7 @@
...
 import pip
 from setuptools import setup
 from setuptools import Command
```
ajaust, line 7: Is this an old version of the `setup.py`? Where are the `try` blocks etc.?

versioneer.py
```diff
@@ -0,0 +1,1855 @@
...
+==============
+
+* like a rocketeer, but for versions!
+* https://github.com/python-versioneer/python-versioneer
+* Brian Warner
+* License: Public Domain
+* Compatible with: Python 3.6, 3.7, 3.8, 3.9 and pypy3
+* [![Latest Version][pypi-image]][pypi-url]
+* [![Build Status][travis-image]][travis-url]
+
```
ajaust, line 16: I this a file from `py-versioneer`? If so, are we allowed to bundle it with our code?

ajaust (commented):
I have added some comments here. I think my biggest concern is the complexity of the solution. It seems like a lot of changes and code (including new dependencies) for getting the `__version__` property.

cyprecice/cyprecice.pyx
BenjaminRueth: This is just internal naming. See https://github.com/precice/python-bindings/blob/fix_pkg/precice/__init__.py. Here the python package `precice` imports everything from `cyprecice`.

setup.py
BenjaminRueth: https://github.com/precice/python-bindings/pull/63 is not merged yet. You mean these `try` blocks, I assume?

versioneer.py
BenjaminRueth: As far as I understand how the versioneer works yes. I followed the install instructions under https://github.com/python-versioneer/python-versioneer/blob/master/INSTALL.md and this file got generated when installing the versioneer.

precice/_version.py
BenjaminRueth: Good point. I will try this out later.

versioneer.py
ajaust: I read quickly through it. As far as I understand `python-versioneer` creates these scripts here. Does it qualify as a dependency in the `pyproject.toml `? If I understand it correctly, we deliver the module `versioneer` via the `versioneer.py` so we do not rely on `python-versioneer` to be installed?!

setup.py
ajaust: Ah yes, that is what I meant. I guess I am just too used to the "new" `setup.py.`

versioneer.py
BenjaminRueth: Correct. I removed the dependency in 2a28881 and uninstalled `versioneer` from my system. Everything is still working.

setup.py
BenjaminRueth: I don't think so. Especially, if we use `versioneer.py`. I quickly skimmed through the file and do not have the impression that the functionalty of `packaging.version` (concatenation/comparison of versions) is provided here.

pyproject.toml
BenjaminRueth: Fixed in 2a28881.

BenjaminRueth:
> I think my biggest concern is the complexity of the solution.

@ajaust I totally agree. To me it also seems a bit too much just for getting `__version__`. But I could not find a simple solution for directly adding `__version__` in the cython package. Some ideas to make this PR a bit less overwhelming:

* I could move the "restructuring" into a different PR. Basically 449abab + some minor things. I think that this is in any case a step into the right direction, since this makes the repo a bit more tidy.
* I think we cannot avoid introducing the wrapper python package (I could not find a pure cython way)
* We could avoid using the versioneer. There are also more minimalistic approaches (see https://stackoverflow.com/a/7071358/5158031). Even though I think that the versioneer has some nice features and will also help us to streamline the release cycle.

Any opinions on the suggestions above?

ajaust:
@BenjaminRueth I am much more supporting of the changes now some more explanations regarding `versioneer` and the additional changes. My biggest fear to have a new (build) dependency just for this feature, but since it is not needed and since  you are convinced of `versioneer` I would suggest to stick with it. Especially since you see further potential for streamlining processes due to `versioneer`. If we become unhappy with the situation, we can still remove `versioneer` and go for a different approach.

So I am fine with using `versioneer` and the restructuring. Whether you prefer to do the restructuring here or in a different PR I leave to you. I think you know better what to do than I do.

precice/_version.py
BenjaminRueth: I checked the `precice.__version__` on a system without git and it worked. This was a bit surprising, since in the source-code git is actually called. But the versioneer is doing something clever here. Looking at `$HOME/.local/lib/python3.8/site-packages/precice/_version.py` I found out that the version is hard-coded and no git is needed anymore by the user. Looks good :smile:

BenjaminRueth:
> So I am fine with using `versioneer` and the restructuring. Whether you prefer to do the restructuring here or in a different PR I leave to you. I think you know better what to do than I do.

I'm doing the restructuring in #71. After that we can merge #70.

IshaanDesai (approved):
The solution looks quite complicated but purely from the user perspective the bindings work as expected.

BenjaminRueth:
I'm done now with moving all the non-related changes out of this PR (#71 and f5de714). The diff should be much clearer now. I will merge this PR into develop soon.

PR #73, opened 2020-11-23
Fix bug: It looks like opal_init failed for some reason

### LINKED ISSUE #72: Bug: It looks like opal_init failed for some reason
Reported in https://github.com/precice/systemtests/issues/262.

The error is actually an "old friend": https://github.com/precice/precice/issues/296.

Can be fixed following https://github.com/precice/precice/issues/296#issuecomment-467802979

My first careful guess is that this problem was probably reintroduced in https://github.com/precice/python-bindings/commit/f5de714567eb0b886e4bf93d03878f63cc90ffe4#diff-929be0dcb483bab373f1c5874bd371ce15db5a655bf2cfafeecea32102bb7af4L11.

BenjaminRueth:
Closes #72.

`from mpi4py import MPI` has to be imported in the `pyx`, not in the `pxd` file.

### DIFF
--- a/cyprecice/cyprecice.pxd
+++ b/cyprecice/cyprecice.pxd
@@ -6,7 +6,6 @@ The python module precice offers python language bindings to the C++ coupling li
 cimport numpy as np
 cimport cython
 cimport SolverInterface
-from mpi4py import MPI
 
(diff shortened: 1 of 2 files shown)


BenjaminRueth:
I verified that this PR fixes the bug by running https://github.com/precice/systemtests/blob/c9dcb90ed2f33a1990039e936ac4c63442b31898/tests/Test_bindings/Dockerfile#L22 on this branch.

Issue #79, opened 2021-01-15
Usability: Calling write/read_block functions with float data


BenjaminRodenberg:
A user just ran into the following error:
```
Traceback (most recent call last):
  File "/home/benjamin/python-bindings/test/test_bindings_module.py", line 198, in test_read_write_block_scalar_data_single_float
    solver_interface.write_block_scalar_data(1, 1, write_data)
  File "cyprecice/cyprecice.pyx", line 954, in cyprecice.Interface.write_block_scalar_data
    if len(values) > 0:
TypeError: object of type 'int' has no len()
```
I was able to reproduce this behavior (will push the branch in a moment) by providing a single float as `write_data` to `write_block_scalar_data`.

### Main question

Is this wrong API usage? I think the C++ preCICE API would not even compile with this kind of input. What do we want to do in the python world? The code obviously runs, but the error message is not helpful.

### Possible Solutions

Depending on the answer to the question above:

* Do we want to raise an error, because it is wrong API usage?
* Do we want to raise a warning, if we that the user is using the API in the wrong way and pass the arguments to `write_scalar_data`?
* Do we want to call the "right" function automatically depending on user input? Maybe completely entirely drop the "block" vs. "non-block", "scalar" vs. "vector" interface and call the right function depending on user input (just `read_data` and `write_data`)? We can implement this in a non-breaking fashion. This is the most pythonic, but also most radical solution, since it really deviates from the C++ API. Maybe this is a good "experimental" feature?

BenjaminRodenberg:
We can start working on this issue on the branch [`i_79`](https://github.com/precice/python-bindings/tree/i_79) as soon as we have concluded on the original question.

IshaanDesai:
> Do we want to raise an error, because it is wrong API usage?

According to me this is the natural step to fix this immediately. But then there are many such possible API usages where error messages are necessary. This is on-going work here: https://github.com/precice/python-bindings/pull/9 (This PR has been pending for a long time, needs to be completed in the near future :see_no_evil: )

> Do we want to call the "right" function automatically depending on user input? Maybe completely entirely drop the "block" vs. "non-block", "scalar" vs. "vector" interface and call the right function depending on user input (just read_data and write_data)? We can implement this in a non-breaking fashion. This is the most pythonic, but also most radical solution, since it really deviates from the C++ API. Maybe this is a good "experimental" feature?

This is the best solution and easily possible with the python bindings. Upvote to do this at some point in the future :+1:

[...]

PR #80, opened 2021-01-15
Check that input is array_like, if required

BenjaminRodenberg:
Related to #79.

### DIFF
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -451,11 +451,8 @@ cdef class Interface:
         """
         check_array_like(position, "position", "set_mesh_vertex")
 
-        if not isinstance(position, np.ndarray):
-            position = np.asarray(position)
-
         if len(position) > 0:
-            dimensions = position.size
+            dimensions = len(position)
             assert(dimensions == self.get_dimensions())
         elif len(position) == 0:
             dimensions = self.get_dimensions()
@@ -584,9 +581,6 @@ cdef class Interface:
         """
         check_array_like(vertex_ids, "vertex_ids", "get_mesh_vertices")
 
-        if not isinstance(vertex_ids, np.ndarray):
-            positions = np.asarray(vertex_ids)
-
         cdef np.ndarray[int, ndim=1] _vertex_ids = np.ascontiguousarray(vertex_ids, dtype=np.int32)
         size = _vertex_ids.size
         cdef np.ndarray[double, ndim=1] _positions = np.empty(size * self.get_dimensions(), dtype=np.double)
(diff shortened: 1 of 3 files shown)


BenjaminRodenberg:
@MakisH @IshaanDesai Note: I will implement `check_array_like` for all functions that actually accept a `array_like`, if you agree with the general approach. Just don't want to roll this out to the whole bindings just to revert/change it. Let's prototype this first at the example of `read_block_scalar_data` and `write_block_scalar_data`

BenjaminRodenberg:
You can see the actual error here: https://github.com/precice/python-bindings/pull/80/checks?check_run_id=1710161099#step:7:122

MakisH (approved):
This error message is very helpful and much better than before! :+1: 

I have not tried running the code. The target branch looks correct. Maybe squash your two commits.

BenjaminRodenberg:
Now all API functions are fixed in the same fashion. Waiting for the review of @IshaanDesai. I will squash everything into one single commit :+1:

cyprecice/cyprecice.pyx
```diff
@@ -568,6 +582,11 @@ cdef class Interface:
         >>> positions.shape
         (5, 3)
         """
+        check_array_like(vertex_ids, "vertex_ids", "get_mesh_vertices")
+
+        if not isinstance(vertex_ids, np.ndarray):
```
IshaanDesai, line 587:
```suggestion
        if not isinstance(positions, np.ndarray):
```

cyprecice/cyprecice.pyx
```diff
@@ -568,6 +582,11 @@ cdef class Interface:
...
+            positions = np.asarray(vertex_ids)
```
IshaanDesai, line 588: Is this necessary?

IshaanDesai (approved):

cyprecice/cyprecice.pyx
BenjaminRodenberg: Good eye! It's not. `np.ascontiguousarray` already takes care of converting everything that we get into a `np.ndarray`.

BenjaminRodenberg: Removed this line and tests still work. The test `test_set_mesh_vertex_list` should be testing exactly this case.

BenjaminRodenberg: Took care of this case an similar ones in c065016d0fe622e8b5c4de9754289316c7c046dd. 

Some insights: Converting to a numpy array is only neaded, of `input.shape` is required before converting `input` via `np.ascontiguousarray`. This is only the case for vector valued input (vector data and positions), where we want to make sure that number of dimensions and samples fit with the corresponding ids.

PR #85, opened 2021-02-24
ValueError FAQ

oguzziya:


README.md
```diff
@@ -268,6 +268,20 @@ python setup.py install --user
 ```
 </details>
 
+### ValueError while importing preCICE
+If you face with an error something like
```
BenjaminRodenberg, line 272:
```suggestion
If you face the error:
```

README.md
```diff
@@ -268,6 +268,20 @@ python setup.py install --user
...
+
+```bash
+ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject
+```
+
+make sure that you are using up-to-date version of NumPy. You can update NumPy with
```
BenjaminRodenberg, line 278:
```suggestion
make sure that you are using an up-to-date version of NumPy. You can update NumPy with
```

README.md
```diff
@@ -268,6 +268,20 @@ python setup.py install --user
...
+
+```bash
+pip3 install numpy --upgrade
```
BenjaminRodenberg, line 281: Shouldn't we add the `--user` flag here? Or does `pip3` magically do the right thing when upgrading? Did you have to run this command as `sudo`?

BenjaminRodenberg (approved):
Generally looks good. Just some minor corrections and a question.

README.md
oguzziya: My knowledge of pip is limited, unfortunately. When I checked it up I could not find any practices to follow. Adding`--user` would be the safe option I think. I did not use `sudo` while running it but in general I never had to use sudo while using pip.

BenjaminRodenberg: Then let's just keep it as it is. I think newer pip version apply the `--user` flag by default.

PR #89, opened 2021-03-03
Overhaul spack CI

BenjaminRodenberg:
This PR simplifies the spack CI.

* Uses approach proposed in https://github.com/haampie-spack/ci-example/commit/b98a5e8b1fb7e8aec67707a7eced78efa92b0619.
* Uses Cron job for creating base image
* Automatically gets dependencies from `package.py` using `--only=dependencies`
* Using jinja and template is not required anymore, since we now use `dev-build` to build the package directly from the branch

Thanks again @haampie for your talk at SIAM CSE 21! Feel free to add comments or a review to this PR, if you want.


BenjaminRodenberg:
https://github.com/precice/python-bindings/pull/89/commits/74a5dbfad53a25b33bb37791fa1920464e246053 might break the tests, but this should be resolved as soon as the upstream PR https://github.com/spack/spack/pull/22059 arrives here.

fsimonis:
Why do we need the local `py-pyprecice` spack package? Isn't [this](https://github.com/spack/spack/blob/49c015e3919db47ba21c441e135ae832b8656c95/var/spack/repos/builtin/packages/py-pyprecice/package.py#L21) enough for this pipeline?

haampie:
I usually add add a copy of `package.py` to the sources so that I can change the build script, add dependencies, etc in a pull request and test it immediately. When merged I create a PR to upstream spack with the new package.py. The downside is that you manually have to sync it with upstream changes to your package.

ajaust:
@fsimonis I think the idea was to have a CI pipeline that one can run on any branch and not just `develop`..

I personally like(d) the approach using a `package.py` such that one can set up a proper environment with root spec. I can see that the (old) approach using the template to generate a `package.py` that references the branch you actually want to test had its advantages. However, our workflow has not been optimal as we also build a new `package.py` more or less by hand for merging it into the Spack repository. Maybe it would have been worth extending the template approach to generate new `package.py` for ci and for updating the actual Spack package?

I am no expert on Spack's `dev-build` feature. If it makes life much easier it could be worth it after all. However, I also read some discussion about either the `dev-build` or `build-dev` command of Spack where it was not clear whether the command will be removed at some point or not. I am not sure what the status on that is at the moment. I could not find the discussion either.

.github/workflows/build-spack.yml
```diff
@@ -8,33 +8,18 @@ jobs:
...
-      - name: Extract branch name  # from https://stackoverflow.com/a/58035262/5158031
-        shell: bash
-        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
-        id: extract_branch
-      - name: Specify Package Script
+          cp spack/repo/packages/py-pyprecice/package.py /sources/spack/repo/packages/py-pyprecice/
+          cp spack/repo/packages/py-pyprecice/*.patch /sources/spack/repo/packages/py-pyprecice/
+      - name: Try to build py-pyprecice with spack and test it
         run: |
-          python3 spack/jinja-instantiate.py --branch ${{ steps.extract_branch.outputs.branch }} > /opt/spack/var/spack/repos/builtin/packages/py-pyprecice/package.py
```
ajaust, line 31: Do we (still) need the explicit `openssh` dependency? I never had to add or compile it on any of the computers that I have used the package on.

spack/ci-spack-pyprecice-deps-1804.dockerfile
```diff
@@ -1,8 +1,10 @@
...
-    && spack install
+FROM spack/ubuntu-bionic:latest 
+WORKDIR /sources
+# Mount the current sources into the build container
+# and build the default environment
+ADD . /sources
+RUN spack --color=always env create --without-view ci /sources/spack/spack.yaml && \
+    spack --color=always -e ci repo add /sources/spack/repo && \
+    spack --color=always -e ci install --fail-fast --only=dependencies && \
+    spack --color=always clean -a
```
ajaust, line 10: I  think we have to evaluate here what is better. The approach have specs in the environment or the new one. Why do you create the environment `ci` using `--without-view ci`?

spack/repo/packages/py-pyprecice/package.py
```diff
@@ -18,7 +18,7 @@ class PyPyprecice(PythonPackage):
     maintainers = ["ajaust", "BenjaminRueth"]
```
ajaust, line 18: Update `BenjaminRueth` to `BenjaminRodenberg`

spack/repo/packages/py-pyprecice/package.py
```diff
@@ -18,7 +18,7 @@ class PyPyprecice(PythonPackage):
...
 
     # Always prefer final version of release candidate
-    version("develop", branch="{{ branch }}")
```
ajaust, line 21: Not needed anymore, but I guess `package.py` should have been called `package.py_template` or so in earlier versions.

spack/spack.yaml
```diff
@@ -0,0 +1,6 @@
+spack:
+  specs:
+    - py-pyprecice@develop
```
ajaust, line 3: Is there a risk that it will be misinterpreted because `@develop` might **not** point to the actual develop branch?

ajaust (commented):
i added my reviews. I think its basically the question whether `dev-build` is better than using the template-generated `package.py`.

BenjaminRodenberg:
> @fsimonis I think the idea was to have a CI pipeline that one can run on any branch and not just `develop`..

Exactly. I think this is a nice feature, since it allows us to test compatibility of a feature with spack before we merge it into develop.

If we take this feature as a "must", then I really prefer the `dev-build` over the jinja template solution for the following reason:

* I think it is more simple (of course subjective)
* it allows to use the same `package.py` for submitting to the upstream spack repository and testing
* I was in the end to remove quite some documentation [here](https://github.com/precice/python-bindings/pull/89/files#diff-795f4dd3dbca2157b8dcb98e63869e9706ddcdf9860e472c1f2b2335dc798d2dL5-L8)
* I could also remove a lot of code required by jinja [in the gitlab action workflows](https://github.com/precice/python-bindings/pull/89/files?file-filters%5B%5D=.yml#diff-9b2263421cd80adc0670965840ba077cb11c8b3dfca837dcf8017d84be6982eeL19-L31)

spack/spack.yaml
BenjaminRodenberg: I think for our testing this "misinterpretation" is exactly what we want: `dev-build` uses the current version of an arbitrary branch as a patch for `py-pyprecice@develop`. Then we install/use (the patched) `py-pyprecice@develop` in all our tests. Does this make sense? If yes: @ajaust feel free to mark this as resolved. If no: Let's try to get to a conclusion here. I should probably also document this somewhere...

spack/repo/packages/py-pyprecice/package.py
BenjaminRodenberg: Yes. This would have avoided some confusion in the past.

spack/repo/packages/py-pyprecice/package.py
BenjaminRodenberg: I think this change is already applied upstream in https://github.com/spack/spack/pull/22012. I will take care of synchronizing upstream with `python-bindings/develop`, then merge `develop` into this PR. This should resolve this point.

spack/ci-spack-pyprecice-deps-1804.dockerfile
BenjaminRodenberg: I copied this approach from https://github.com/haampie-spack/ci-example/blob/b98a5e8b1fb7e8aec67707a7eced78efa92b0619/tools/docker/build-env.Dockerfile#L7-L12. So I cannot really give a good reason. I'm not sure, if this is really a satisfactory answer, but I'll try to get through the review with it ;)

.github/workflows/build-spack.yml
BenjaminRodenberg: From my point of view: yes. I added it to solve a runtime dependency of OpenMPI. See https://github.com/precice/python-bindings/pull/76. @ajaust feel free to mark this conversation as resolved, if this is satisfactory.

spack/repo/packages/py-pyprecice/package.py
BenjaminRodenberg: Resolved by https://github.com/precice/python-bindings/pull/89/commits/8ee7c7665f7538d99d481ebf9c067063501718db

spack/spack.yaml
BenjaminRodenberg: I removed the `spack.yaml` in e3371fd. I think this new approach is clearer (I hope it works as expected).

BenjaminRodenberg:
Currently the test fails due to changes in bad16b0. As soon as [this job](https://github.com/precice/python-bindings/runs/2033029098?check_suite_focus=true) is done, I will rerun the test and it is expected to succeed.

.github/workflows/build-spack.yml
BenjaminRodenberg: I just realized that `openssh` was removed in d2216f6ff9e74635c93d09a04313e58f8cf8d06a and reintroduced it in 5d725fdb09108ff8acb153229a647a033a9449d0.

BenjaminRodenberg:
Currently the test fails, because [`openssh` is missing](https://github.com/precice/python-bindings/runs/2034682880?check_suite_focus=true#step:5:2011) on the baseimage. [This job](https://github.com/precice/python-bindings/runs/2034635313?check_suite_focus=true) is currently rebuilding the baseimage including `openssh`, but this will take a bit longer.

fsimonis:
> I think the idea was to have a CI pipeline that one can run on any branch and not just develop.

With `dev-build` you can run the pipeline on every branch as develop: _"Dear spack, please install this package at the given spec, but use the sources in this directory instead of downloading them."_

.github/workflows/build-spack.yml
```diff
@@ -9,32 +9,17 @@ jobs:
...
-          ls spack/var/spack/repos/builtin/packages/py-pyprecice
-          cp spack/var/spack/repos/builtin/packages/py-pyprecice/*.patch /opt/spack/var/spack/repos/builtin/packages/py-pyprecice/
-          ls spack/var/spack/repos/builtin/packages/py-pyprecice
-      - name: Try to build py-pyprecice with spack
-        run: |
-          . /opt/spack/share/spack/setup-env.sh && spack env activate precice && spack add py-pyprecice@develop ^precice@develop && spack concretize -f && spack install -v
-      - name: Import precice and print version
-        run: |
-          . /opt/spack/share/spack/setup-env.sh && spack env activate precice && python3 -c "import precice; print(precice.__version__)"
+          . /opt/spack/share/spack/setup-env.sh && spack env activate ci && spack arch && spack find && spack dev-build py-pyprecice@develop target=x86_64 && spack load precice py-numpy py-mpi4py py-cython openssh openmpi && mkdir runner && cd runner && python3 -c "import precice; print(precice.__version__)"
```
BenjaminRodenberg, line 25: @fsimonis We are using `dev-build` here. Is this what you mean?

ajaust (approved):
Looks good to me know. I assume the `dev-build` discussion is (somewhat) resolved for the moment.

haampie:
Fwiw, you might also be interested in https://github.com/spack/spack/pull/22115

Issue #90, opened 2021-03-04
Solver dummy partly out of date


ajaust:
The solver dummy seems to be out to date, see  https://github.com/precice/python-bindings/tree/develop/solverdummy. It tells users to install the requirements to via
```
pip3 install --user -r requirements.txt
```
which seems to fail, see https://precice.discourse.group/t/python-bindings-error-solver-dummy/502/3

The `requirements.txt` expects `pyprecice==2.0`. Isn't this too strict and should be `pyprecice>=2.0`? Do we need the further version digits, i.e., `pyprecice==2.0.0.0`?

BenjaminRodenberg:
@ajaust If I get it right #91 will close this issue?

ajaust:
Yes, this issued is closed by #91 which has been merged.

ajaust:
I was too fast. It has not been merged yet. Feel free to close this issue after #91 has been merged.

BenjaminRodenberg:
> I was too fast. It has not been merged yet. Feel free to close this issue after #91 has been merged.

You are so fast. Nobody would have seen it anyway :rocket:

PR #91, opened 2021-03-04
Update requirements.txt

ajaust:
Allow for `pyprecice` versions that are newer than 2.0. A quick check on my local machine worked out. 

That is related to https://github.com/precice/python-bindings/issues/90

Closes https://github.com/precice/python-bindings/issues/90

### DIFF
--- a/solverdummy/requirements.txt
+++ b/solverdummy/requirements.txt
@@ -1,3 +1,3 @@
-pyprecice==2.0
+pyprecice>=2.0
 argparse>=1.4
 numpy>=1.16


BenjaminRodenberg (approved):
Thanks! Looks good to me. I also tried installing from `develop` as well as from this branch and observed the same you reported in https://github.com/precice/python-bindings/issues/90.

You suggested adding the missing digits in the version number. I.e. `pyprecice>=2.0.0.0` instead of `pyprecice>=2.0`. I think that this is optional. We can even write `pyprecice>=2`, if we want. I don't have a opinion here. I'm fine with any of the solutions above. Feel free to merge this PR either as it is or with a four or single digit version number.

ajaust:
Then I will leave it as `2.0` for now. If we find out that we do not like it anymore, we can still change it in the future.

PR #97, opened 2021-04-07
Make release guide more generic

IshaanDesai:
This PR proposes that the `ReleaseGuide.md` be made more generic and should not have a specific version number in its description.

### DIFF
--- a/docs/ReleaseGuide.md
+++ b/docs/ReleaseGuide.md
@@ -1,25 +1,25 @@
 ## Guide to release new version of python-bindings
 The developer who is releasing a new version of the python-bindings is expected to follow this work flow:
 
-The release of the `python-bindings` repository is made directly from a release branch called `python-bindings-v2.1.1.1`. This branch is mainly needed to help other developers with testing.
+The release of the `python-bindings` repository version `2.x.y.z` is made directly from a release branch called `python-bindings-v2.x.y.z`. This branch is mainly needed to help other developers with testing.
 
-1. Create a branch called `python-bindings-v2.1.1.1` from the latest commit of the `develop` branch.
+1. Create a branch called `python-bindings-v2.x.y.z` from the latest commit of the `develop` branch.
 
-2. [Open a Pull Request `master` <-- `python-bindings-v2.1.1.1`](https://github.com/precice/python-bindings/compare/master...master) named after the version (i.e. `Release v2.1.1.1`) and briefly describe the new features of the release in the PR description.
+2. [Open a Pull Request `master` <-- `python-bindings-v2.x.y.z`](https://github.com/precice/python-bindings/compare/master...master) named after the version (i.e. `Release v2.x.y.z`) and briefly describe the new features of the release in the PR description.
 
 3. Bump the version in the following places:
 
-    * `CHANGELOG.md` on `python-bindings-v2.1.1.1`.
+    * `CHANGELOG.md` on `python-bindings-v2.x.y.z`. For this change pushing directly to the release branch is permissible.
     * There is no need to bump the version anywhere else, since we use the [python-versioneer](https://github.com/python-versioneer/python-versioneer/) for maintaining the version everywhere else.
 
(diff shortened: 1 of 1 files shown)


docs/ReleaseGuide.md
```diff
@@ -1,25 +1,25 @@
...
     * There is no need to bump the version anywhere else, since we use the [python-versioneer](https://github.com/python-versioneer/python-versioneer/) for maintaining the version everywhere else.
 
-4. [Draft a New Release](https://github.com/precice/python-bindings/releases/new) in the `Releases` section of the repository page in a web browser. The release tag needs to be the exact version number (i.e.`v2.1.1.1` or `v2.1.1.1rc1`, compare to [existing tags](https://github.com/precice/python-bindings/tags)). Use `@target:master`. Release title is also the version number (i.e. `v2.1.1.1` or `v2.1.1.1rc1`, compare to [existing releases](https://github.com/precice/python-bindings/tags)). 
+4. [Draft a New Release](https://github.com/precice/python-bindings/releases/new) in the `Releases` section of the repository page in a web browser. The release tag needs to be the exact version number (i.e.`v2.1.1.1` or `v2.1.1.1rc1`, compare to [existing tags](https://github.com/precice/python-bindings/tags)). Use `@target:master`. Release title is also the version number (i.e. `v2.x.y.z` or `v2.x.y.zrc1`, compare to [existing releases](https://github.com/precice/python-bindings/tags)).
```
BenjaminRodenberg, line 15: I find the notation `v2.x.y.zrc1` a bit confusing.

BenjaminRodenberg, line 15:
```suggestion
4. [Draft a New Release](https://github.com/precice/python-bindings/releases/new) in the `Releases` section of the repository page in a web browser. The release tag needs to be the exact version number (i.e.`v2.x.y.z` or `v2.x.y.zrc1`, compare to [existing tags](https://github.com/precice/python-bindings/tags)). Use `@target:master`. Release title is also the version number (i.e. `v2.x.y.z` or `v2.x.y.zrc1`, compare to [existing releases](https://github.com/precice/python-bindings/tags)). 
```

BenjaminRodenberg: Additionally: It should be `vw.x.y.zrca` to be consistent (which is totally unreadable). Why not use capital letters? ` vW.X.Y.ZrcA` but the `W` and `A` are a bit strange. Probably due to our non-standard versionin scheme...

BenjaminRodenberg (changes requested):
I see some inconsistencies. Additionally: I'm not sure if this is really needed. In the end `v2.1.1.1` is a more specific and easier to read example. Of course, `v2.1.1.1` is already outdated, but for the sake of an example it's fine. Why do you think we should make the release guide more generic? Is the current example not ok?

IshaanDesai:
> I see some inconsistencies. Additionally: I'm not sure if this is really needed. In the end `v2.1.1.1` is a more specific and easier to read example. Of course, `v2.1.1.1` is already outdated, but for the sake of an example it's fine. Why do you think we should make the release guide more generic? Is the current example not ok?

I am mainly thinking from the perspective of future developers who might be confused why a release guide is referring to a specific version number. Another solution would be to add one sentence at the top:

*The version v2.1.1.1 is used as an example version number in this release guide*

Does this sound more reasonable?

BenjaminRodenberg:
> > I see some inconsistencies. Additionally: I'm not sure if this is really needed. In the end `v2.1.1.1` is a more specific and easier to read example. Of course, `v2.1.1.1` is already outdated, but for the sake of an example it's fine. Why do you think we should make the release guide more generic? Is the current example not ok?
> 
> I am mainly thinking from the perspective of future developers who might be confused why a release guide is referring to a specific version number. Another solution would be to add one sentence at the top:
> 
> _The version v2.1.1.1 is used as an example version number in this release guide_
> 
> Does this sound more reasonable?

The generic version looks a bit confusing to me, since we have a non standard versioning and some more not so common things (`rc` for release candidate). Combining this with letters instead of number seems a bit too much to me. I you think there is a danger of confusion in the future this sentence might help.

Note: If `v2.1.1.1` is an ancient version in a few years and this looks just strange, we can still update it to `v4.5.6.7`, if we want. Of course, not on every release, since this is just unnecessary work.

IshaanDesai:
Discussion has led to the conclusion that making the release guide more generic makes it more confusing and hence these changes are not to be finalized.

PR #101, opened 2021-04-23
Remove mpi4py as dependency

BenjaminRodenberg:
This PR removes mpi4py (again) as a dependency.

There is some history and some other issues involved. I just added a test in #100 to be able to reproduce possible failures. Before merging this PR I would like to investigate the setup described in https://github.com/precice/python-bindings/issues/8#issuecomment-575329609 again and make sure that the error is really gone (ideally also find out why it is now gone).

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,6 @@ Please refer to [precice.org](https://www.precice.org/installation-bindings-pyth
 
 **preCICE**: Refer to [the preCICE wiki](https://github.com/precice/precice/wiki#1-get-precice) for information on building and installation.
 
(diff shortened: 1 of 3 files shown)


BenjaminRodenberg:
I don't have the time to carefully check the MPI dependency now. I'm still a bit afraid here that we might oversee something, if we merge this PR, but only having MPI as an optional dependency would still be nice. I would suggest to keep this issue. If somebody finds time to work on it or the MPI dependency really becomes a problem (because one cannot easily satisfy it on a special system, for example), we can start looking into this again.

PR #102, opened 2021-04-27
Adding GitHub Actions to run partitioned-heat-conduction tutorial cases

IshaanDesai:
This PR adds GitHub Actions scripts to run the tutorial cases `partitioned-heat-conduction` and `partitioned-heat-conduction-complex`. No verification is done, just successful running of the tutorials is tested.

### DIFF
--- a/.github/workflows/run-tutorials.yml
+++ b/.github/workflows/run-tutorials.yml
@@ -80,5 +80,5 @@ jobs:
         run: git clone -b develop https://github.com/precice/tutorials.git
       - name: Run tutorial
         run: | 
-          cd tutorials/partitioned-heat-conduction/fenics
+          cd tutorials/partitioned-heat-conduction-complex/fenics
           python3 heat.py -d -i complex & python3 heat.py -n -i complex
--- a/.github/workflows/run-tutorials.yml
+++ b/.github/workflows/run-tutorials.yml
@@ -1,0 +1,84 @@
+name: Run preCICE Tutorials
+on:
+  push:
+    branches:
+      - master
+      - develop
+  pull_request:
+    paths:
+      - '**'
+      
+jobs:  
+  run_ht_simple:
+    name: Run partitioned-heat-conduction-simple
+    runs-on: ubuntu-latest
+    container: precice/precice
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


BenjaminRodenberg (changes requested):
I'm a bit unsure whether the `python-bindings` are the right place for this kind of test. The test depends on the `fenics-adapter`, which is a downstream component from the perspective of the bindings. We may, of course, ask the question "If I change the python-bindings, does this break the installation of the fenics-adapter? Are there breaking API changes?", but I think it is not common that a dependency checks whether code that depends on it is still working.

Another problem I see with this testing procedure: If there is a breaking change, how do you want to fix this test? The test will break, since the upstream component does not implement the breaking change. But it does not make sense to update the upstream component, before the breaking change is released (releasing, even though a test is not passing).

From this perspective I would rather see this test in

1. the `fenics-adapter`: we already have it [here](https://github.com/precice/fenics-adapter/blob/develop/.github/workflows/run-tutorials.yml), but should use [`precice/python-bindings`](https://github.com/precice/python-bindings/blob/develop/.github/workflows/build-docker.yml) as base image to get rid of the boilerplate for installing the bindings.
2. the `tutorials` or `systemtests`: similar to the test above, but using [`precice/fenics-adapter`](https://github.com/precice/fenics-adapter/blob/develop/.github/workflows/build-docker.yml) as base image to get rid of the boilerplate for installing the bindings and the adapter.

MakisH:
I agree that we should keep every upstream component agnostic of what is happening downstream. This should be caught by the system tests (i.e. tutorials / cron job).

Imagine the following simpler example: the preCICE library itself would not directly check whether a change affects the FEniCS adapter and try to adapt to the adapter. The adapter would try to adapt to preCICE. Similarly, the Python bindings should check that they still work with preCICE, and the FEniCS adapter should check if it still works with the Python bindings.

Or, even further: the Ubuntu developers should not worry if preCICE still works on Ubuntu. :-) But definitely, we should be checking our projects in the system tests.

IshaanDesai:
> the question "If I change the python-bindings, does this break the installation of the fenics-adapter? Are there breaking API changes?"

This was my rationale behind adding this test here in the bindings. I agree with all the concerns raised above and do see how this test is better suited to be in `fenics-adapter` or `tutorials` or best case `systemtests`.

PR #103, opened 2021-04-29
Remove Travis CI

IshaanDesai:
This PR removes Travis CI as the mock tests run in the CI are now run as a GitHub Actions [job](https://github.com/precice/python-bindings/blob/develop/.github/workflows/build-and-test.yml#L71-L96)


BenjaminRodenberg (changes requested):
Removing this test is good. As far as I undestand the corresponding github actions test is this one:

https://github.com/precice/python-bindings/blob/6e54396f393a5243d3514d36be1a6ad48c8c3127/.github/workflows/build-and-test.yml#L71-L96

I still see a minor problem here: The old test on travis uses a generic docker image as basis (probably some ubuntu?). The github-actions-based test uses [`precice/ci-ubuntu-1804`](https://github.com/precice/ci-images/blob/master/ci-ubuntu-1804.dockerfile). If possible, please use ubuntu:20.04 or a similarly simple image as basis, since otherwise this contradicts the idea of mocking a bit. You can use the [`fenics-adapter` mock test](https://github.com/precice/fenics-adapter/blob/b248cf013e976e67f6d3bed5335a943cee2542e0/.github/workflows/build-and-test.yml#L28-L44) as a reference. Here, `container: quay.io/fenicsproject/stable:current` provides `fenics`, but all the other dependencies of preCICE and preCICE itself can be ignored and are therefore not present.



Here, https://github.com/precice/fenics-adapter/blob/b248cf013e976e67f6d3bed5335a943cee2542e0/.github/workflows/build-and-test.yml#L31

IshaanDesai:
I see the problem with using the container `precice/ci-ubuntu-1804`, it does indeed violate the broad mock testing principle. But in the `test/SolverInterface.cpp` we do import the original preCICE header file here:
https://github.com/precice/python-bindings/blob/6e54396f393a5243d3514d36be1a6ad48c8c3127/test/SolverInterface.cpp#L1

In my understanding this requires the container to have the preCICE compiled or at least the code be present there. Is this correct? If the GitHub action is run on just a ubuntu container then the following error is seen:
```
| /home/desaiin/python-bindings/cyprecice/cyprecice.cpp:671:10: fatal error: precice/SolverInterface.hpp: No such file or directory
|   671 | #include "precice/SolverInterface.hpp"
|       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
| compilation terminated.
| error: command 'x86_64-linux-gnu-gcc' failed with exit status 1
[Build and Test/Run setup test]   ❌  Failure - Run setup test
```

BenjaminRodenberg:
> I see the problem with using the container `precice/ci-ubuntu-1804`, it does indeed violate the broad mock testing principle. But in the `test/SolverInterface.cpp` we do import the original preCICE header file here:
> 
> https://github.com/precice/python-bindings/blob/6e54396f393a5243d3514d36be1a6ad48c8c3127/test/SolverInterface.cpp#L1
> 
> In my understanding this requires the container to have the preCICE compiled or at least the code be present there. Is this correct? If the GitHub action is run on just a ubuntu container then the following error is seen:
> 
> ```
> | /home/desaiin/python-bindings/cyprecice/cyprecice.cpp:671:10: fatal error: precice/SolverInterface.hpp: No such file or directory
> |   671 | #include "precice/SolverInterface.hpp"
> |       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> | compilation terminated.
> | error: command 'x86_64-linux-gnu-gcc' failed with exit status 1
> [Build and Test/Run setup test]   ❌  Failure - Run setup test
> ```

`precice/SolverInterface.hpp` is needed. Yes. But the rest of preCICE is not needed. The idea here is the following:

* Use the real `precice/SolverInterface.hpp` to ensure that the interface that we are testing is the real preCICE interface
* Use a fake [`SolverInterface.cpp`](https://github.com/precice/python-bindings/blob/develop/test/SolverInterface.cpp) as implementation.

The important part is here:

https://github.com/precice/python-bindings/blob/6e54396f393a5243d3514d36be1a6ad48c8c3127/.github/workflows/build-and-test.yml#L84-L88

We clone precice, since we need `precice/SolverInterface.hpp` and then move it to a location, where the compiler can discover it (`precice/SolverInterface.hpp`). preCICE itself is never installed, all other header files are missing.

If we then want to build test, we do not link against the real preCICE (it does not exist on the system), but just provide the fake `SolverInterface.cpp` as sources:

https://github.com/precice/python-bindings/blob/6e54396f393a5243d3514d36be1a6ad48c8c3127/setup.py#L72-L76

I hope this explanation helps. It's not documented well and I assume must be very confusing, if you don't know where you have to look.

.github/workflows/build-and-test.yml
```diff
@@ -71,25 +71,24 @@ jobs:
...
         uses: actions/checkout@v2
-      - name: Install & upgrade pip3
+      - name: Install & upgrade pip3 and install OpenMPI
         run: |
-          apt-get -yy update
-          apt-get install -y python3-pip
-          rm -rf /var/lib/apt/lists/*
+          sudo apt-get -yy update
+          sudo apt-get install -y python3-pip libopenmpi-dev
+          sudo rm -rf /var/lib/apt/lists/*
```
BenjaminRodenberg, line 81: Related to #101: We should be able to remove the mpi part here.

But for the current state everything is fine.

.github/workflows/build-and-test.yml
```diff
@@ -71,25 +71,24 @@ jobs:
...
-          cp precice-core/src/precice/SolverInterface.hpp precice/SolverInterface.hpp
       - name: Install dependencies
         run: |
           pip3 install --user toml
           python3 -c 'import toml; c = toml.load("pyproject.toml"); print("\n".join(c["build-system"]["requires"]))' | pip3 install -r /dev/stdin
+      - name: Checkout precice and make required files discoverable 
+        run: |  
+          git clone https://github.com/precice/precice.git precice-core 
+          mkdir -p precice 
+          cp precice-core/src/precice/SolverInterface.hpp precice/SolverInterface.hpp
```
BenjaminRodenberg, line 91: Is there a reason for moving `Checkout precice and make required files discoverable` down? If not, I would keep it where it is for a cleaner diff.

BenjaminRodenberg (approved):
Looks good. I will quickly check and apply the minor things I've spotted and then directly merge this PR.

.github/workflows/build-and-test.yml
BenjaminRodenberg: Done in 0ef71e3.

.github/workflows/build-and-test.yml
IshaanDesai: Okay, will keep this dependency problem in mind.

PR #104, opened 2021-04-30
Install python and pip via action.

BenjaminRodenberg:


BenjaminRodenberg:
The [pip action](https://github.com/marketplace/actions/pip-installer) seems to not work with custom containers (i.e. `precice/precice`). This means its only relevant for the test `setup_test` and `pythonpublish`, since all other tests are using `precice/precice`.

BenjaminRodenberg:
The [test run](https://github.com/precice/python-bindings/runs/2478559817) was successful. The upload was not successful, since there was no tag provided. But all the building and uploading worked.

IshaanDesai (approved):

Issue #86, opened 2021-03-03
Phase install_lib fails using Spack if system's Python is used


ajaust:
i tried to install the bindings using Spack on a system where external packages were set up with `spack external find`. This command also found the pre-installed Python of the system. Now, building the package `py-pyprecice` fails with the following error message:

```
==> Installing py-pyprecice-2.2.0.1-z4ojdrz6aqadiqidezkkubiifvwjr4uk
==> No binary for py-pyprecice-2.2.0.1-z4ojdrz6aqadiqidezkkubiifvwjr4uk found: installing from source
==> Fetching https://github.com/precice/python-bindings/archive/v2.2.0.1.tar.gz
##################################################################################################################################### 100.0%#-=#=- #     #                                                                                                                             
==> No patches needed for py-pyprecice
==> py-pyprecice: Executing phase: 'install_lib'
==> Error: ProcessError: Command exited with status 1:
    '/usr/bin/python3.9' '-s' 'setup.py' '--no-user-cfg' 'install_lib'

1 error found in build log:
     63     8611 |   __pyx_v_self->thisptr->writeBlockScalarData(__pyx_t_4, __pyx_v_size, ((int const *)__pyx_v__vertex_ids->data), ((doubl
           e const *)__pyx_v__values->data));
     64          |   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     65    /home/alex/software/spack/lib/spack/env/gcc/g++ -shared -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -fno-semantic-interpos
           ition -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now build/temp.linux-x86_64-3.9/tmp/alex/spack-stage/spack-stage-py-pyprecic
           e-2.2.0.1-z4ojdrz6aqadiqidezkkubiifvwjr4uk/spack-src/cyprecice/cyprecice.o -L/usr/lib -o build/lib.linux-x86_64-3.9/cyprecice.cp
           ython-39-x86_64-linux-gnu.so -lprecice
     66    UPDATING build/lib.linux-x86_64-3.9/precice/_version.py
[...]

PR #105, opened 2021-05-18
Release v2.2.1.1

IshaanDesai:
This release is required due to a [bug fix release of preCICE v2.2.1](https://github.com/precice/precice/pull/1023)

Features of this release:
* Removing Travis CI
* Adding tests to check autopep8 formatting, building dockerimage


IshaanDesai:
@BenjaminRodenberg is the Spack build expected to fail right now?

BenjaminRodenberg (approved):
From my side this PR looks good.

The failing spack test is a tricky issue. I assume that this is a problem on precice@develop and not on precice@2.2.1. However, we currently don't know, since we do not test using precice@2.2.1. From my perspective a manual test with precice@2.2.1 should be sufficient for this release. On the long run we should think about our testing strategy

1. for the python bindings, when the spack environment is updated (weekly cron job)
2. for precice, since we currently don't have any information on the status of the spack package precice@develop.

ajaust:
The failing Spack build might be simply a Spack artifact when using the Python module in a fresh Spack environment. However, the Spack environment is not touched/changed.

It could also be related https://github.com/precice/python-bindings/issues/86 which means that the Python module is installed to a location where we do not expect/want it to be. Maybe the module is in the environment of a/the wrong Python installation. At least the output suggest that the Python modules was built and installed successfully.

ajaust:
I just restarted the failing test twice and now it passed twice. I am not sure what to conclude here. I would guess this PR is fine for now.

PR #111, opened 2021-07-12
Remove documentation for parallel version

### LINKED ISSUE #110: Solverdummy does not run in parallel
The [description of the solverdummy ](https://github.com/precice/python-bindings/blob/develop/solverdummy/README.md) claims that the solvers can be run in parallel:

>For the parallel version of the solverdummy, you have to add the command mpirun -n <N> to the commands from above.

However, the actual [solverdummy.py](https://github.com/precice/python-bindings/blob/develop/solverdummy/solverdummy.py) has hard coded ranks and communicator size

```python
solver_process_index = 0
solver_process_size = 1

interface = precice.Interface(participant_name, configuration_file_name,
                              solver_process_index, solver_process_size)
```

I am not sure what would be the prepared option:
1. Change the description
2. Change the solverdummy.

I have a parallel version of the dummy in my repository over here: https://github.com/ajaust/precice-parallel-solverdummies

BenjaminRodenberg:
Closes #110

Removes the documentation for the parallel version of the solver dummy. This is consistent with other solverdummies (https://github.com/precice/precice/blob/develop/examples/solverdummies/README.md and https://github.com/precice/fortran-module/blob/develop/examples/solverdummy/README.md). The solverdummies should be as simple as possible and therefore supporting parallel runs is out-of-scope.

### DIFF
--- a/solverdummy/README.md
+++ b/solverdummy/README.md
@@ -10,10 +10,6 @@ You can test the dummy solver by coupling two instances with each other. Open tw
  * `python3 solverdummy.py precice-config.xml SolverOne MeshOne`
  * `python3 solverdummy.py precice-config.xml SolverTwo MeshTwo`
 
-## parallel version
-
-For the parallel version of the solverdummy, you have to add the command `mpirun -n <N>` to the commands from above.
-
 # Next Steps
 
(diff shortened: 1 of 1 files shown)


uekerman:
I am confused, why "only" removing the documentation? Does the solverdummy support parallel runs or not?

uekerman (approved):

Issue #87, opened 2021-03-03
Consider replacing build_lib by build_py phase in Spack recipe


ajaust:
We could check whether it is possible to replace the phase `build_lib` with `build_py` in the Spack recipe as it creates the needed `_version.py`. However, it fails with an error since the directory `build/lib.linux-x86_64-3.9/precice/` seems to be missing.  A grep of the logfile is given here:
```
11:copying precice/_version.py -> build/lib/precice                                                                                         
13:UPDATING build/lib/precice/_version.py                                                                                                   
14:set build/lib/precice/_version.py to 'v2.2.0.1'                                                                                          
72:UPDATING build/lib.linux-x86_64-3.9/precice/_version.py                                                                                  
73:error: [Errno 2] No such file or directory: 'build/lib.linux-x86_64-3.9/precice/_version.py'
```

BenjaminRodenberg:
@ajaust connected to #113 and #86: Could you give a short update here?

ajaust:
Yes, I missed this issue. It is fixed by https://github.com/spack/spack/pull/25077

PR #112, opened 2021-07-19
Update cmdclass in setup.py to allow standard installation routine

ajaust:
I am currently trying to fix https://github.com/precice/python-bindings/issues/86 and https://github.com/precice/python-bindings/issues/87. Instead of trying to use another build phase I tried to go back to the initial setup of a `build_ext` phase and an `install` phase when installing the bindings via Spack. 

I could actually achieve this by restructuring the `setup.py` and the defintiion of the `cmdclass`. I update the internal command classes and this works for me. Together with my Spack package it

- Creates a `_version.py`
- Installs the bindings in the directory specified by Spack

A Spack fork using my fork of the Python bindings can be found [here](https://github.com/ajaust/spack/tree/fix-installation-routine-py-pyprecice). It contains a new package `py-pyprecice-fork` for installing that fork.

I am not completely sure of the implications of my changes. Thus, I would need some feedback whether this change of `setup.py` is legal and whether it breaks any of our tests or use cases.


ajaust:
It also seems that I did something wrong when pulling upstream... :(

Should I close this PR, rebase proberly and open a new PR?

BenjaminRodenberg (approved):
Hi Alex,

thanks for looking into this! I currently don't see any changes that worry me. The PR is luckily quite small. If it works and our tests succeed I think we should try this out. Currently our tests already cover the most important use-cases (I hope).

For me it is extremely hard to tell whether a setup.py works for all our use-cases as intended or not, therefore I'm mostly relying on the tests. If we later find out that this PR broke something, we definitely should add another test (but that's a different story).

I installed the bindings locally with `pip3 install --user .` and it worked. Also `python3 -c "import precice; print(precice.__version__)"` worked as expected.

BenjaminRodenberg:
> It also seems that I did something wrong when pulling upstream... :(
> 
> Should I close this PR, rebase proberly and open a new PR?

Since the changes are very few, this is probably the easiest solution. Please just trigger me again for a review and I will quickly do another review on the new PR.

ajaust:
Great! I am currently playing around with this `setup.py` and Spack including a patch for older versions of the bindings. So far it seems to work as expected. I will do a bit more testing and then create a new PR for the python bindings and I will prepare a PR for our Spack recipe.

ajaust:
I have created a new PR at https://github.com/precice/python-bindings/pull/113 so I will close this PR.

PR #113, opened 2021-07-22
Update cmdclass in setup.py to allow standard installation routine (clean fork)

ajaust:
This is almost the same as https://github.com/precice/python-bindings/pull/112, but from a clean fork. I also added a comment to the `setup.py` about updating the `cmdclass` object of Versioneer.

**Below follows the message in the original PR:**
I am currently trying to fix https://github.com/precice/python-bindings/issues/86 and https://github.com/precice/python-bindings/issues/87. Instead of trying to use another build phase I tried to go back to the initial setup of a `build_ext` phase and an `install` phase when installing the bindings via Spack. 

I could actually achieve this by restructuring the `setup.py` and the defintiion of the `cmdclass`. I update the internal command classes and this works for me. Together with my Spack package it

- Creates a `_version.py`
- Installs the bindings in the directory specified by Spack

A Spack fork using my fork of the Python bindings can be found [here](https://github.com/ajaust/spack/tree/fix-installation-routine-py-pyprecice). It contains a new package `py-pyprecice-fork` for installing that fork.

I am not completely sure of the implications of my changes. Thus, I would need some feedback whether this change of `setup.py` is legal and whether it breaks any of our tests or use cases.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -127,7 +127,7 @@ this_directory = os.path.abspath(os.path.dirname(__file__))
 with open(os.path.join(this_directory, 'README.md'), encoding='utf-8') as f:
     long_description = f.read()
 
-
+# Update versioneer's cmdclass to respect such that it builds the bindings
 cmdclass = versioneer.get_cmdclass()
 cmdclass.update(build_ext=my_build_ext)
 cmdclass.update(install=my_install)
--- a/setup.py
+++ b/setup.py
@@ -127,16 +127,17 @@ this_directory = os.path.abspath(os.path.dirname(__file__))
 with open(os.path.join(this_directory, 'README.md'), encoding='utf-8') as f:
     long_description = f.read()
 
-my_cmdclass = {
-    'test': my_test,
-    'build_ext': my_build_ext,
-    'install': my_install}
+
+cmdclass = versioneer.get_cmdclass()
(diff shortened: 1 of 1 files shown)


BenjaminRodenberg (approved):
Looks good to me. It's always a bit difficult to be aware of all the implications. But we have some tests for the most important use-cases and if they work, I think we don't have to worry.

I find it confusing that creating a `dict` and handing it to `versioneer.get_cmdclass(my_dict)` and your proposed implementation have different outcome. Maybe we should open an issue upstream?

ajaust:
I think we can drop this PR for now. I installed `py-precice@2.2.1.1` with the current `develop` of Spack which contains the changes of https://github.com/spack/spack/pull/25077 and installation worked like a charm. This package does **not** include the changes I have proposed here anymore.

I assume what I have observed was some side effect of an older Spack version or so. At least I have seen significant higher success rate when installing packages via Spack using the system's Python for the current `develop` branch.

BenjaminRodenberg:
I'll close this PR, since it is not needed anymore. See above. https://github.com/precice/python-bindings/issues/86 is already closed without the content of this PR. For #87, status is still unclear, but we can reopen this PR, if related to #87.

PR #116, opened 2021-08-11
Update Spack recipe

ajaust:
This PR updates the Spack recipe included in the repository:

The Spack recipe included in the repository is updated to the current version which is included in the official Spack repository. There have been significant changes and simplifications in a recent PR https://github.com/spack/spack/pull/25077.

I think it makes sense to keep the Spack recipe in our repository up to date.

Replaces #115


spack/repo/packages/py-pyprecice/package.py
```diff
@@ -15,11 +15,13 @@ class PyPyprecice(PythonPackage):
     homepage = "https://www.precice.org"
     git = "https://github.com/precice/python-bindings.git"
     url = "https://github.com/precice/python-bindings/archive/v2.0.0.1.tar.gz"
-    maintainers = ["ajaust", "BenjaminRodenberg"]
+    maintainers = ["ajaust", "BenjaminRodenberg", "IshaanDesai"]
```
BenjaminRodenberg, line 18: Already confirmed in https://github.com/precice/python-bindings/pull/115#discussion_r685818194.

spack/repo/packages/py-pyprecice/package.py
```diff
@@ -28,35 +30,25 @@ class PyPyprecice(PythonPackage):
...
-            self.setup_py("install", "--prefix={0}".format(prefix))
+    depends_on("py-cython@0.29:", type="build")
+    depends_on("py-packaging", when="@:2.1", type="build")
+    depends_on("py-pip", when="@:2.1", type="build")
+
+    @when("@:2.1")
+    def patch(self):
+        filter_file(
+            "distutils.command.install", "setuptools.command.install", "setup.py"
+        )
```
BenjaminRodenberg, line 54: Does this change allow us to remove `spack/repo/packages/py-pyprecice/deactivate-version-check-via-pip.patch` without breaking older releases? If yes: Nice! If we are breaking older releases: No problem from my perspective, but we should document it in the Changelog.

BenjaminRodenberg (approved):
Since this PR mostly reflects https://github.com/spack/spack/pull/25077, I think we should merge it without minor changes. For the future: Let's just synchronize with the upstream spack recipe without any PRs. If there are additional changes that we want to apply on top of the upstream recipi, I would suggest to independently develop this here and merge it into the upstream spack repository with the next release.

Only thing we should do: Add a Changelog entry here, if there a major changes.

spack/repo/packages/py-pyprecice/package.py
ajaust: Yes, it does.  As mentioned in https://github.com/spack/spack/pull/25077 I have tested this with `py-pyprecice@2.0.0.1`, `py-pyprecice@2.1.1.2` and `pyprecice@2.2.1.1` which represents all installation procedures we had so far. 

- `2.0.0.1` is the initial installation process.
- In `2.0.0.2:2.1.1.2` there were some slight changes to the `setup.py`. 
- In `2.2.0.1:` versioneer was introduced which represents the current installation routine as far as I know.

BenjaminRodenberg: Great. Then this PR should be ready to be merged :rocket:

BenjaminRodenberg:
I'm reverting the contribution of this PR in https://github.com/precice/python-bindings/pull/121/. Will reopen a new PR with the contents of this PR in a moment.

PR #117, opened 2021-08-11
Update Spack CI 

ajaust:
This PR updates the Spack based CI included in the repository:

- I have renamed the namespace of the self-defined Spack repository in spack/repo/ to pyprecice.test such that it is easier to distinguish from the Spack commands.
- I updated the Dockerfile to add our own repository before py-pyprecice is added to the environment.
- I updated the Dockerfile and the CI configuration to explicitly use `pyprecice.text.py-pyrecice@develop`.
  - Here I would need feedback if that makes sense for the Build Spack workflow or if we should still base it on the "normal" `py-pyprecice` included in Spack.

Replaces #115


BenjaminRodenberg (approved):
Looks good :+1: 
Just a minor addition: Added a changelog entry.

ajaust:
Is it possible to run the action here to see that it did not break?

BenjaminRodenberg:
Currently the action does not work for two reasons:

### 1. Action not triggered

Action is not triggered by PR (fixed by 9cfdb5e). We should add this change directly on `develop`.

### 2. Forks/Branches not properly supported

`precice/ci-spack-pyprecice-deps-1804` is still the "old" container based on the [dockerfile from `develop`](https://github.com/precice/python-bindings/blob/0d9b589d0a6b3fd16ab7872436230ecea9efd4d4/spack/ci-spack-pyprecice-deps-1804.dockerfile). But we need the [dockerfile from this PR](https://github.com/ajaust/python-bindings/blob/update/spack-ci/spack/ci-spack-pyprecice-deps-1804.dockerfile). Currently it's not easily possible to work with branches/forks here, because the tag of the dockerfile is hardcoded:

https://github.com/precice/python-bindings/blob/0d9b589d0a6b3fd16ab7872436230ecea9efd4d4/.github/workflows/build-spack.yml#L14

https://github.com/precice/python-bindings/blob/0d9b589d0a6b3fd16ab7872436230ecea9efd4d4/.github/workflows/build-env.yml#L28

We should actually push to `forks-docker-id/ci-spack-pyprecice-deps-1804` and use this image, instead.

ajaust:
If you are fine with that, we can merge and keep this PR in mind so we know where the problem should be.

BenjaminRodenberg:
Currently build spack is failing (as expected). See https://github.com/precice/python-bindings/runs/3308743649?check_suite_focus=true.

I just triggered building the environment manually via the `workflow_dispatch`. The job will take some time (https://github.com/precice/python-bindings/actions/runs/1122874035). As soon as the job has run through, the container should be updated and we can use it in the test. If we rerun the test, it should succeed.

ajaust:
I am curious to see how that will work out. :)

BenjaminRodenberg:
I've rebuilt the environment and the test fails now: https://github.com/precice/python-bindings/runs/3320630602?check_suite_focus=true. So there is a bug we have to fix.

BenjaminRodenberg:
This PR seems to be fine, but #116 is problematic. See https://github.com/precice/python-bindings/pull/121/.

PR #118, opened 2021-08-11
Run CI also on PRs

BenjaminRodenberg:
Resulting from https://github.com/precice/python-bindings/pull/117, since tests are currently not triggered by PRs from forks.


ajaust (approved):
Looks good to me so far.

PR #121, opened 2021-08-13
Revert #116 and fix #117

BenjaminRodenberg:
The failure of the test in https://github.com/precice/python-bindings/commit/36e4782b8460ba93ec6bfc307bef7bf281c540fa seems to be caused by the additions from #116. Actually https://github.com/precice/python-bindings/commit/c5b19c786ad2b527dea08efe0e895c68df3ed3e9 was already failing. The missing `patch` was hiding another bug. Therefore I'm reverting #116 while keeping #117 in this PR.


BenjaminRodenberg:
I just created https://github.com/precice/python-bindings/pull/122 to revise the contribution that this #121 will remove.

PR #122, opened 2021-08-13
Update Spack recipe (bugfix)

BenjaminRodenberg:
This PR is an update of https://github.com/precice/python-bindings/pull/116, which has been/will be reverted via https://github.com/precice/python-bindings/pull/121


IshaanDesai:
Is this PR still relevant?

ajaust:
I think it can be removed since we have a new(er) version of the Spack package in the repository at the moment. However, I am not sure if the initial cause of this PR ([failing test](https://github.com/precice/python-bindings/pull/121)) has been fixed or whether the test is not used anymore?!

PR #124, opened 2021-09-08
Adding direct mesh access experimental functions

### LINKED ISSUE #123: Add API functions related to direct access to received meshes
In https://github.com/precice/precice/pull/1033 new API functions were added to enable direct access to received meshes. The new API functions are:
* [setMeshAccessRegion](https://github.com/precice/precice/blob/fb6e44baf35db4d8baac2dc49f8c99cf875cd91f/src/precice/SolverInterface.hpp#L838)
* [getMeshVerticesAndIDs](https://github.com/precice/precice/blob/fb6e44baf35db4d8baac2dc49f8c99cf875cd91f/src/precice/SolverInterface.hpp#L854)

IshaanDesai:
Adding two new API functions related to functionality for direct access to received meshes. More details: https://github.com/precice/precice/pull/1033
closes #123

### DIFF
--- a/cyprecice/SolverInterface.pxd
+++ b/cyprecice/SolverInterface.pxd
@@ -105,9 +105,9 @@ cdef extern from "precice/SolverInterface.hpp" namespace "precice":
 
         # direct mesh access
 
-        void setMeshAccessRegion (const int meshID, const double* boundingBox)
+        void setMeshAccessRegion (const int meshID, const double* boundingBox) const
 
-        void getMeshVerticesAndIDs (const int meshID, const int size, int* ids, double* coordinates)
+        void getMeshVerticesAndIDs (const int meshID, const int size, int* ids, double* coordinates) const
 
 cdef extern from "precice/SolverInterface.hpp" namespace "precice":
     string getVersionInformation()
--- a/cyprecice/SolverInterface.pxd
+++ b/cyprecice/SolverInterface.pxd
@@ -101,6 +101,12 @@ cdef extern from "precice/SolverInterface.hpp" namespace "precice":
 
         void readScalarData (const int dataID, const int valueIndex, double& value)
 
+        # direct mesh access
+
+        void setMeshAccessRegion (const int meshID, const double* boundingBox)
+
+        void getMeshVerticesAndIDs (const int meshID, const int size, int* ids, double* coordinates)
+
 cdef extern from "precice/SolverInterface.hpp" namespace "precice":
     string getVersionInformation()
 
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -8,6 +8,7 @@ The python module precice offers python language bindings to the C++ coupling li
 cimport cyprecice
 import numpy as np
 from mpi4py import MPI
+import warnings
 
 from cpython.version cimport PY_MAJOR_VERSION  # important for determining python version in order to properly normalize string input. See http://docs.cython.org/en/latest/src/tutorial/strings.html#general-notes-about-c-strings and https://github.com/precice/precice/issues/68 .
 
@@ -1195,22 +1196,21 @@ cdef class Interface:
 
     def set_mesh_access_region (self, mesh_id, bounding_box):
         """
-        EXPERIMENTAL
         This function is required if you don't want to use the mapping schemes in preCICE, but rather 
         want to use your own solver for data mapping. As opposed to the usual preCICE mapping, only a
         single mesh (from the other participant) is now involved in this situation since an 'own' 
         mesh defined by the participant itself is not required any more. In order to re-partition the
         received mesh, the participant needs to define the mesh region it wants read data from and 
         write data to. The mesh region is specified through an axis-aligned bounding box given by the
-        lower and upper [min and max] bounding-box limits in each space dimension [x, y, z].
+        lower and upper [min and max] bounding-box limits in each space dimension [x, y, z]. This function is still
+        experimental
 
         Parameters
         ----------
         mesh_id : int
             ID of the mesh you want to access through the bounding box
         bounding_box : array_like
-            Axis aligned bounding boxes which has in 3D the format 
-            [x_min, x_max, y_min, y_max, z_min, z_max]
+            Axis aligned bounding box. Example for 3D the format: [x_min, x_max, y_min, y_max, z_min, z_max]
 
         Notes
         -----
@@ -1237,6 +1237,8 @@ cdef class Interface:
         0.5, i.e. the defined access region as computed through the involved provided mesh is by 50% 
         enlarged.
         """
+        warnings.warn("The function set_mesh_access_region is still experimental.")
+
         check_array_like(bounding_box, "bounding_box", "set_mesh_access_region")
         cdef np.ndarray[double, ndim=1] _bounding_box = np.ascontiguousarray(bounding_box, dtype=np.double)
 
@@ -1245,7 +1247,7 @@ cdef class Interface:
     def get_mesh_vertices_and_ids (self, mesh_id):
         """
         Iterating over the region of interest defined by bounding boxes and reading the corresponding 
-        coordinates omitting the mapping.
+        coordinates omitting the mapping. This function is still experimental.
 
         Parameters
         ----------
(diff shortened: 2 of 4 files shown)


IshaanDesai:
All tests fail because the docker image `precice/precice` is made from the `master` branch of the preCICE repo and the new API functions only exist on the `develop` branch at this point of time.

cyprecice/cyprecice.pyx
```diff
@@ -1193,6 +1193,81 @@ cdef class Interface:
         self.thisptr.readScalarData (data_id, vertex_id, _value)
         return _value
 
+    def set_mesh_access_region (self, mesh_id, bounding_box):
+        """
+        EXPERIMENTAL
```
BenjaminRodenberg, line 1198: SInce we used the numpy documentation as a template, you can also check here: https://github.com/numpy/numpy/blob/2d112a98ed7597c4120b31908384ae09b0304659/numpy/array_api/__init__.py#L1-L125

Some observations:

* experimental state is only briefly mentioned in the docs.
* a warning is raised.

cyprecice/cyprecice.pyx
```diff
@@ -1193,6 +1193,81 @@ cdef class Interface:
...
+        the safety factor region resulting from the mapping. The default value of the safety factor is 
+        0.5, i.e. the defined access region as computed through the involved provided mesh is by 50% 
+        enlarged.
+        """
+        check_array_like(bounding_box, "bounding_box", "set_mesh_access_region")
+        cdef np.ndarray[double, ndim=1] _bounding_box = np.ascontiguousarray(bounding_box, dtype=np.double)
+
+        self.thisptr.setMeshAccessRegion(mesh_id, <double*>_bounding_box.data)
+
+    def get_mesh_vertices_and_ids (self, mesh_id):
```
BenjaminRodenberg, line 1245: Also experimental?

cyprecice/cyprecice.pyx
```diff
@@ -1193,6 +1193,81 @@ cdef class Interface:
...
+        received mesh, the participant needs to define the mesh region it wants read data from and 
+        write data to. The mesh region is specified through an axis-aligned bounding box given by the
+        lower and upper [min and max] bounding-box limits in each space dimension [x, y, z].
+
+        Parameters
+        ----------
+        mesh_id : int
+            ID of the mesh you want to access through the bounding box
+        bounding_box : array_like
+            Axis aligned bounding boxes which has in 3D the format
```
BenjaminRodenberg, line 1212:
```suggestion
            Axis aligned bounding box. Example for 3D format:
```

cyprecice/cyprecice.pyx
```diff
@@ -1193,6 +1193,81 @@ cdef class Interface:
...
+        write data to. The mesh region is specified through an axis-aligned bounding box given by the
+        lower and upper [min and max] bounding-box limits in each space dimension [x, y, z].
+
+        Parameters
+        ----------
+        mesh_id : int
+            ID of the mesh you want to access through the bounding box
+        bounding_box : array_like
+            Axis aligned bounding boxes which has in 3D the format 
+            [x_min, x_max, y_min, y_max, z_min, z_max]
```
BenjaminRodenberg, line 1213: Why not the same format as for vertices? So: `n x d`, where `n = 2`, because you have two vertices `v_min` and `v_max`

cyprecice/cyprecice.pyx
```diff
@@ -1193,6 +1193,81 @@ cdef class Interface:
...
+
+        The safety factor scaling (see safety-factor in the configuration file) is not applied to the 
+        defined access region and a specified safety will be ignored in case there is no additional
+        mapping involved. However, in case a mapping is in addition to the direct access involved, you
+        will receive (and gain access to) vertices inside the defined access region plus vertices inside
+        the safety factor region resulting from the mapping. The default value of the safety factor is 
+        0.5, i.e. the defined access region as computed through the involved provided mesh is by 50% 
+        enlarged.
+        """
+        check_array_like(bounding_box, "bounding_box", "set_mesh_access_region")
```
BenjaminRodenberg, line 1240: I would suggest to check the shape of the `bounding_box` here. We do a similar check for input vertices, if I remember correctly. You should, for example ensure that for 2d the array has a different shape than for 3d.

cyprecice/cyprecice.pyx
BenjaminRodenberg: I got a bit carried away here: https://github.com/precice/precice/issues/1111

test/SolverInterface.cpp
```diff
@@ -49,6 +51,7 @@ SolverInterface::SolverInterface(
   n_fake_vertices = 3;
   fake_ids.resize(n_fake_vertices);
   std::iota(fake_ids.begin(), fake_ids.end(), 0);
+  fake_bounding_box = std::vector<double>();
```
BenjaminRodenberg, line 54: Currently this is not needed.

test/SolverInterface.cpp
```diff
@@ -350,6 +353,20 @@ void SolverInterface:: readScalarData
     value = fake_read_write_buffer[0]; 
 }
 
+void SolverInterface:: setMeshAccessRegion
+(
+  int meshID,
+  const double* boundingBox ) const
+{}
```
BenjaminRodenberg, line 360: You could implement a test, where you input a bounding box (e.g. `[1,2,3,4,5,6]`) via the python bindings and ensure here (on the cpp side) that the bounding box is received correctly. 

Same for the `meshID`: Input `0` on the python side and ensure that `0` also arrives here.

test/SolverInterface.cpp
```diff
@@ -350,6 +353,20 @@ void SolverInterface:: readScalarData
...
+
+void SolverInterface:: getMeshVerticesAndIDs
+(
+  int meshID,
+  int size,
+  int* valueIndices,
+  double* coordinates ) const
+{}
```
BenjaminRodenberg, line 368: Here, you could generate some dummy values on the cpp side and check whether you receive them correctly on the python side.

test/SolverInterface.cpp
```diff
@@ -350,6 +353,20 @@ void SolverInterface:: readScalarData
...
+void SolverInterface:: setMeshAccessRegion
+(
+  int meshID,
+  const double* boundingBox ) const
+{}
+
+void SolverInterface:: getMeshVerticesAndIDs
+(
+  int meshID,
+  int size,
```
BenjaminRodenberg, line 365:
```suggestion
  const int meshID,
  const int size,
```
?

test/SolverInterface.cpp
```diff
@@ -350,6 +353,20 @@ void SolverInterface:: readScalarData
     value = fake_read_write_buffer[0]; 
 }
 
+void SolverInterface:: setMeshAccessRegion
+(
+  int meshID,
```
BenjaminRodenberg, line 358:
```suggestion
  const int meshID,
```

test/SolverInterface.cpp
```diff
@@ -350,6 +353,20 @@ void SolverInterface:: readScalarData
...
+  const double* boundingBox ) const
+{}
+
+void SolverInterface:: getMeshVerticesAndIDs
```
BenjaminRodenberg, line 362:
```suggestion
void SolverInterface:: getMeshVerticesAndVertexIDs
```
Before it sounded like "get mesh vertices and mesh ids" now it sounds like "get mesh vertices and vertex ids". I think this is what is happening.

BenjaminRodenberg (changes requested):
I added some more comments below. Some smaller issues on documentation. Testing is probably a slightly bigger work package.

There are also some more comments that probably should go to the C++ side of the interface. But I think you can judge better here. I'm missing a lot of context about the C++ interface, so they might not 100% fit.

test/SolverInterface.cpp
IshaanDesai: Okay, I will open an issue in the upstream repo. Lets stick to what we have for now.

cyprecice/SolverInterface.pxd
```diff
@@ -101,6 +101,12 @@ cdef extern from "precice/SolverInterface.hpp" namespace "precice":
 
         void readScalarData (const int dataID, const int valueIndex, double& value)
 
+        # direct mesh access
+
+        void setMeshAccessRegion (const int meshID, const double* boundingBox)
+
+        void getMeshVerticesAndIDs (const int meshID, const int size, int* ids, double* coordinates)
```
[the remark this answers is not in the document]
BenjaminRodenberg: Currently the code compiles, so it is no real problem. But I would still add `const` consistently with the C++ API for consistency. Please add `const` here, if there is a `const` in the C++ API. For the remaining functions where a `const` is missing: We should clean this up independent of this PR.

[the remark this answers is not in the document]
IshaanDesai: I made the new API functions `const` and also opened another PR to make other functions consistent https://github.com/precice/python-bindings/pull/128

PR #126, opened 2021-09-14
Add API function to access mesh connectivity requirement

IshaanDesai:
Adding API function `is_mesh_connectivity_required()` based on changes to original API here: https://github.com/precice/precice/pull/1080

### DIFF
--- a/cyprecice/SolverInterface.pxd
+++ b/cyprecice/SolverInterface.pxd
@@ -77,6 +77,8 @@ cdef extern from "precice/SolverInterface.hpp" namespace "precice":
 
         # data access
 
+        bool isMeshConnectivityRequired (int meshID)
+
         bool hasData (const string& dataName, int meshID) const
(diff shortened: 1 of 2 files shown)


IshaanDesai:
@BenjaminRodenberg I am not sure why the `setup_install` and subsequent workflows are failing even though the function `is_mesh_connectivity_required()` exists in the right places. Can you please check once?

test/SolverInterface.cpp
```diff
@@ -153,6 +153,13 @@ bool SolverInterface::hasToEvaluateFineModel() const
   return 0;
 }
 
+bool SolverInterface:: isMeshConnectivityRequired
+(
+  int           meshID )
```
BenjaminRodenberg, line 158:
```suggestion
  int           meshID ) const
```
I added a `const` here and now it works.

PR #135, opened 2021-10-21
Use pkg-config to determine flags

### LINKED ISSUE #5: Use pkg-config to detect flags
It may be worth looking into the [pkgconfig](https://pypi.org/project/pkgconfig/) package.

We could use this to extract the necessary compiler flags for preCICE.

BenjaminRodenberg:
Closes #5.

@fsimonis: One thing I'm worrying a bit: What if a user does not have `pkg-config` installed? Or would you just take this as granted?


ajaust:
And another note... If we want to use `pkg-config` we could also think about using the `pkgconfig` package available in Python which would maybe make the setup files look a bit nicer.

IshaanDesai:
The python package `pkgconfig` is now used to determine preCICE flags in the bindings. This was implemented in https://github.com/precice/python-bindings/pull/149

PR #138, opened 2022-03-29
Add Action for MacOS installation

ajaust:
TBD

### DIFF
--- a/.github/workflows/build-and-test-macos.yml
+++ b/.github/workflows/build-and-test-macos.yml
@@ -1,0 +1,61 @@
+name: Build and Test Python bindings on macOS
+on:
+  push:
+    branches-ignore:
+      - '**'
+  workflow_dispatch:
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


ajaust:
The action currently fails as the bindings do not find the preCICE headers. I am not sure whether I should fix it by setting the `CPATH` as this action currently installs preCICE in the `/usr` directory. We should find out, if this should be in the standard search path of MacOS X first.

kursatyurt:
On my computer using miniconda python installation without any problem.  

I just changed CFLAGS in setup.py to include `-L\usr\local\include`

But again I have linker problem

BenjaminRodenberg:
Just merged develop into this PR to keep it synchronized. Seems like there are some problems. I'm not planning to work on this PR since I do not have particular interest in MacOS but would be great if any MacOS user could pick this up.

IshaanDesai:
Keep this PR updated in the hope that we would merge the MacOS Action at some point.

PR #140, opened 2022-03-29
Add C++ compiler as dependency

ajaust:
This mention a C++ compiler as dependency for building the bindings in the README. It was pointed out that installation of the bindings fails, if there is no C++ compiler installed.

Do we actually need MPI as a dependency as well?


IshaanDesai (approved):
Makes sense to have a note that a working C++ compiler needs to exist on the system. Regarding having MPI as a dependency, things are a bit unclear. In the bindings we import `mpi4py` but never really use it, so I am also unsure whether we really need MPI or even `mpi4py` as a dependency. @BenjaminRodenberg am I missing something here?

PR #141, opened 2022-04-22
Moving solverdummy into examples folder

IshaanDesai:
This PR moves the solverdummy to a newly created `examples` folder to make it consistent with the precice repository.


BenjaminRodenberg (approved):
Overall good step into the right direction. From my perspective you can merge this. I just have a few remarks that you can consider, if you want. But the current solution is also fine:

* My muscle memory typed `python3 solverdummy.py ../precice-config.xml SolverTwo MeshTwo`. This, of course, does not work, because the structure is `examples/solverdummy/solvedummy.py` and `precice-config.xml` lies next to `solverdummy.py`, not `examples/solverdummies/python/solverdummy.py` and `precice-config.xml` lies one level above. This would be the 100% consistent version, but also overengineered.
* What about the changes introduced in https://github.com/precice/precice/pull/1256 for the other solverdummies? Could also be done in another PR, but please don't forget it.

IshaanDesai:
I will remove the `meshName` in a different PR so that we do not have file path changes and in-file changes in the same PR.

PR #143, opened 2022-05-17
Fix formatting of Spack documentation and make documentation clearer

ajaust:
This fixes the formatting of the lists in markdown such that they are rendered correctly on GitHub and by other Markdown rendering tools, e.g. Visual Studio Code.

While doing so I also fixed some erroneous commands and rephrased some parts to be more precise.

### DIFF
--- a/spack/README.md
+++ b/spack/README.md
@@ -1,6 +1,6 @@
 # Spack package: py-pyprecice
 
-The Spack package `py-pyprecice` provides the python bindings via Spack and was submitted via https://github.com/spack/spack/pull/19558. This folder contains the Spack package script that can be used for testing whether the Spack installation works.
+The Spack package `py-pyprecice` provides the python bindings via Spack and was submitted via [pull request 19558](https://github.com/spack/spack/pull/19558) in the Spack repository. This folder contains the Spack package script that can be used for testing whether the Spack installation works.
 
 ## Docker image `precice/ci-spack-pyprecice-deps-1804`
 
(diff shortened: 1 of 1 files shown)


spack/README.md
```diff
@@ -10,38 +10,56 @@ The workflow `build_spack` in `.github/workflows/build-spack.yml` uses the image
...
+
   2. Build the new python bindings. Usually, we use a new Spack environment. This is created and loaded via
-  ```
-  spack env create pyprecicetest-VERSIONNUMBER
-  spack env activate -p pyprecicetest-VERSIONNUMBER
-  spack
-  ```
-  Note that the dots `.` in the version number must be replaced by underscores or similar.
+
+     ```text
```
IshaanDesai, line 24: Why is this text sytled when similar commands above are styled bash?

spack/README.md
```diff
@@ -10,38 +10,56 @@ The workflow `build_spack` in `.github/workflows/build-spack.yml` uses the image
...
+     spack env create pyprecicetest-VERSIONNUMBER
+     spack env activate -p pyprecicetest-VERSIONNUMBER
+     ```
+
+     Note that the dots `.` in the version number must be replaced by underscores or similar. This means, one would create an environment named `pyprecicetest-2_0_0_1` instead of `pyprecicetest-2.0.0.1` since the `.` character is not allowed for Spack environments.
```
IshaanDesai, line 29: Can we please describe the *or similar* options or just remove that? I, personally, for example am really new to Spack and would not know other options than `_`. Is `-` also allowed?

spack/README.md
```diff
@@ -10,38 +10,56 @@ The workflow `build_spack` in `.github/workflows/build-spack.yml` uses the image
...
+
   3. We add the `py-pyprecice` package from our repository to the environment via
-  ```
-  spack add pyprecice.test.py-pyprecice@VERSIONNUMBER
-  ```
-  This ensures that one uses the local repository which has the namespace `pyprecice.test` when installing `py-precice`.
+
+     ```text
```
IshaanDesai, line 33: Same as above.

spack/README.md
```diff
@@ -10,38 +10,56 @@ The workflow `build_spack` in `.github/workflows/build-spack.yml` uses the image
...
+     spack add pyprecice.test.py-pyprecice@VERSIONNUMBER
+     ```
+
+     This ensures that one uses the local repository which has the namespace `pyprecice.test` when installing `py-precice`. Note that when speciying the `VERSIONNUMBER` you use the actual version number with dots, e.g. `2.0.0.1`.
```
IshaanDesai, line 37: Is the user not required to mention the full version as `v2.0.0.1`, or is just `2.0.0.1` enough? I have not tried this myself.

spack/README.md
```diff
@@ -10,38 +10,56 @@ The workflow `build_spack` in `.github/workflows/build-spack.yml` uses the image
...
-  ```
-  spack concretize -f
-  ```
-  and check whether everything looks good. You might want to edit the environment of the Spack environment (the `spack.yaml` file) and enforce combined concretization by adding `concretization: together` at the end of the file. You might want to double check whether the correct repository is used by calling
-  ```
-  spack spec -N py-pyprecice@VERSIONNUMBER
-  ```
-  It will print the concretization and prefixes each packages with the repository it will be installed from.
+
+     ```text
```
IshaanDesai, line 41: Same as above

IshaanDesai (approved):
Thanks for the edits and streamlining! Just a few minor comments from my side.

spack/README.md
ajaust: I guess the `bash` was a leftover from old formatting. I homogenized it to `text`.

spack/README.md
ajaust: Yes, `-` is allowed. That is why it is also in the environment name. I rephrased it to be less confusing.

PR #144, opened 2022-05-22
Release v2.4.0.0

IshaanDesai:
Release for [preCICE v2.4.0](https://github.com/precice/precice/releases/tag/v2.4.0)

Main features:

* Move solverdummy into examples/ folder and remove MeshName from its input arguments. https://github.com/precice/python-bindings/pull/141 and https://github.com/precice/python-bindings/pull/142


spack/repo/packages/py-pyprecice/package.py
```diff
@@ -12,14 +12,17 @@ class PyPyprecice(PythonPackage):
     C++ library preCICE.
     """
 
-    homepage = "https://www.precice.org"
+    homepage = "https://precice.org"
     git = "https://github.com/precice/python-bindings.git"
     url = "https://github.com/precice/python-bindings/archive/v2.0.0.1.tar.gz"
```
BenjaminRodenberg, line 17: Why is this specifically `v2.0.0.1`? If it's working ok, but looks strange to me.

IshaanDesai: @ajaust could you please make sure this is correct?

ajaust: It is correct. The `url` property is set to a downloadable archive which is often different from the homepage or git URL. This allows Spack to automatically substitute the version number to any other version number that is defined in the package. Using this substitution, Spack can download the corresponding tar ball. By default, Spack prefers to download archives over `git clone` as one can verify the checksum.

More information is in the Spack documentation on the [Version URLs](https://spack.readthedocs.io/en/latest/packaging_guide.html#version-urls).

spack/repo/packages/py-pyprecice/package.py
```diff
@@ -12,14 +12,17 @@ class PyPyprecice(PythonPackage):
...
-    maintainers = ["ajaust", "BenjaminRodenberg"]
+    maintainers = ["ajaust", "BenjaminRodenberg", "IshaanDesai"]
```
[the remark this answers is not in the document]
ajaust: Seems like we did not import all changes from the offical Spack repository back into the Python adapter repository. @IshaanDesai has been [aboard for a while](https://github.com/spack/spack/blob/c5297523af7387b358511fe15cfb520a6441113c/var/spack/repos/builtin/packages/py-pyprecice/package.py#L18). :smile:

ajaust:
> I did not try to install or use the bindings, but I think we can rely on our CI pipeline.

I did not install it manually either, but I think that the CI looks trustworthy to me. So I would also trust the CI. If there are problems showing up, we should rather improve the CI than testing the bindings manually.

PR #145, opened 2022-07-02
Add gradient related API

kursatyurt:
This PR adds nearest neighbour gradient related API to python bindings

### DIFF
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -1246,7 +1246,7 @@ cdef class Interface:
         >>> data_id = 1
         >>> vertex_ids = [1, 2]
         >>> gradientValues = np.array([[v1x_dx, v1y_dx, v1z_dx, v1x_dy, v1y_dy, v1z_dy, v1x_dz, v1y_dz, v1z_dz], [v2x_dx, v2y_dx, v2z_dx, v2x_dy, v2y_dy, v2z_dy, v2x_dz, v2y_dz, v2z_dz]])
-        >>> interface.write_block_vector_data(data_id, vertex_ids, gradientValues)
+        >>> interface.write_block_vector_gradient_data(data_id, vertex_ids, gradientValues)
         """
         check_array_like(vertex_ids, "vertex_ids", "write_block_vector_gradient_data")
         check_array_like(gradientValues, "gradientValues", "write_block_vector_gradient_data")
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -1215,7 +1215,7 @@ cdef class Interface:
         """
         Writes vector gradient data given as block. This function writes values of specified vertices to a dataID.
         Values are provided as a block of continuous memory. Values are stored in a numpy array [N x D] where N = number
-        of vertices and D = dimensions of geometry squared.
+        of vertices and D = number of gradient components.
 
         Parameters
         ----------
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -1239,7 +1239,7 @@ cdef class Interface:
         Write block gradient vector data for a 2D problem with 2 vertices:
         >>> data_id = 1
         >>> vertex_ids = [1, 2]
-        >>> gradientValues = np.array([[v1x_dx, v1y_dx,v1x_dy, v1y_dy], [v2x_dx, v2y_dx,v2x_dy, v2y_dy]])
+        >>> gradientValues = np.array([[v1x_dx, v1y_dx, v1x_dy, v1y_dy], [v2x_dx, v2y_dx, v2x_dy, v2y_dy]])
         >>> interface.write_block_vector_gradient_data(data_id, vertex_ids, gradientValues)
 
         Write block vector data for a 3D (D=3) problem with 2 (N=2) vertices:
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -1395,7 +1395,7 @@ cdef class Interface:
             
         Examples
         --------
-        Write block gradient vector data for a 2D problem with 2 vertices:
+        Write block gradient scalar data for a 2D problem with 2 vertices:
         >>> data_id = 1
         >>> vertex_ids = [1, 2]
         >>> gradientValues = np.array([[v1_dx, v1_dy],[v2_dx, v2_dy]])
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -1263,7 +1263,7 @@ cdef class Interface:
         cdef np.ndarray[int, ndim=1] _vertex_ids = np.ascontiguousarray(vertex_ids, dtype=np.int32)
         cdef np.ndarray[double, ndim=1] _gradientValues = np.ascontiguousarray(gradientValues.flatten(), dtype=np.double)
 
-        assert _gradientValues.size == size * self.get_dimensions() * self.get_dimensions(), "Vector data is not provided for all vertices in write_block_vector_data. Check length of input data provided. Provided size: {}, expected size: {}".format(_gradientValues.size, size * self.get_dimensions() * self.get_dimensions())
+        assert _gradientValues.size == size * self.get_dimensions() * self.get_dimensions(), "Dimension of vector gradient data provided in write_block_vector_gradient_data does not match problem definition. Check length of input data provided. Provided size: {}, expected size: {}".format(_gradientValues.size, size * self.get_dimensions() * self.get_dimensions())
         assert _vertex_ids.size == size, "Vertex IDs are of incorrect length in write_block_vector_gradient_data. Check length of vertex ids input. Provided size: {}, expected size: {}".format(_vertex_ids.size, size)
 
         self.thisptr.writeBlockVectorGradientData (data_id, size, <const int*>_vertex_ids.data, <const double*>_gradientValues.data)
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -1415,7 +1415,7 @@ cdef class Interface:
 
         if len(gradientValues) > 0:
             size, dimensions = gradientValues.shape
-            assert dimensions == self.get_dimensions() , "Dimensions of vector data in write_block_scalar_gradient_data does not match with dimensions in problem definition. Provided dimensions: {}, expected dimensions: {}".format(dimensions, self.get_dimensions())
+            assert dimensions == self.get_dimensions() , "Dimensions of scalar gradient data  provided in write_block_scalar_gradient_data does not match with dimensions in problem definition. Provided dimensions: {}, expected dimensions: {}".format(dimensions, self.get_dimensions())
         if len(gradientValues) == 0:
             size = 0
 
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -1404,7 +1404,7 @@ cdef class Interface:
         Write block scalar data for a 3D (D=3) problem with 2 (N=2) vertices:
         >>> data_id = 1
         >>> vertex_ids = [1, 2]
-        >>> values = np.array([[v1_dx, v1_dy, v1x_dz], [v2_dx. v2_dy, v2_dz]])
+        >>> values = np.array([[v1_dx, v1_dy, v1x_dz], [v2_dx, v2_dy, v2_dz]])
         >>> interface.write_block_scalar_gradient_data(data_id, vertex_ids, values)
         """
         check_array_like(vertex_ids, "vertex_ids", "write_block_scalar_gradient_data")
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -1427,6 +1427,32 @@ cdef class Interface:
 
         self.thisptr.writeBlockScalarGradientData (data_id, size, <const int*>_vertex_ids.data, <const double*>_gradientValues.data)
 
+    def is_gradient_data_required(self,data_id):
+        """
+        Returns true if gradient data is required for a dataID.
+        Parameters
+        ----------
+        data_id : int
+            Data ID to check.
+
+        Returns
+        -------
+        bool
+            True if gradient data is required for a dataID.
+
+        Notes
+        -----
+        Previous calls:
+            initialize() has been called
+
+        Examples
+        --------
+        Check if gradient data is required for a dataID:
+        >>> data_id = 1
+        >>> interface.is_gradient_data_required(data_id)
+        """
+        return self.thisptr.isGradientDataRequired(data_id)
+
 
     def set_mesh_access_region (self, mesh_id, bounding_box):
         """
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -1240,7 +1240,7 @@ cdef class Interface:
         >>> data_id = 1
         >>> vertex_ids = [1, 2]
         >>> gradientValues = np.array([[v1x_dx, v1y_dx,v1x_dy, v1y_dy], [v2x_dx, v2y_dx,v2x_dy, v2y_dy]])
-        >>> interface.write_block_vector_data(data_id, vertex_ids, gradientValues)
+        >>> interface.write_block_vector_gradient_data(data_id, vertex_ids, gradientValues)
 
         Write block vector data for a 3D (D=3) problem with 2 (N=2) vertices:
         >>> data_id = 1
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -1401,11 +1401,11 @@ cdef class Interface:
         >>> gradientValues = np.array([[v1_dx, v1_dy],[v2_dx, v2_dy]])
         >>> interface.write_block_scalar_gradient_data(data_id, vertex_ids, gradientValues)
 
-        Write block vector data for a 3D (D=3) problem with 2 (N=2) vertices:
+        Write block scalar data for a 3D (D=3) problem with 2 (N=2) vertices:
         >>> data_id = 1
         >>> vertex_ids = [1, 2]
         >>> values = np.array([[v1_dx, v1_dy, v1x_dz], [v2_dx. v2_dy, v2_dz]])
-        >>> interface.write_block_scalar_data(data_id, vertex_ids, values)
+        >>> interface.write_block_scalar_gradient_data(data_id, vertex_ids, values)
         """
         check_array_like(vertex_ids, "vertex_ids", "write_block_scalar_gradient_data")
         check_array_like(gradientValues, "gradientValues", "write_block_sclar_gradient_data")
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -1422,7 +1422,7 @@ cdef class Interface:
         cdef np.ndarray[int, ndim=1] _vertex_ids = np.ascontiguousarray(vertex_ids, dtype=np.int32)
         cdef np.ndarray[double, ndim=1] _gradientValues = np.ascontiguousarray(gradientValues.flatten(), dtype=np.double)
 
-        assert _gradientValues.size == size * self.get_dimensions(), "Vector data is not provided for all vertices in write_block_vector_data. Check length of input data provided. Provided size: {}, expected size: {}".format(_gradientValues.size, size * self.get_dimensions())
+        assert _gradientValues.size == size * self.get_dimensions(), "Scalar gradient data is not provided for all vertices in write_block_scalar_gradient_data. Check length of input data provided. Provided size: {}, expected size: {}".format(_gradientValues.size, size * self.get_dimensions())
         assert _vertex_ids.size == size, "Vertex IDs are of incorrect length in write_block_scalar_gradient_data. Check length of vertex ids input. Provided size: {}, expected size: {}".format(_vertex_ids.size, size)
 
         self.thisptr.writeBlockScalarGradientData (data_id, size, <const int*>_vertex_ids.data, <const double*>_gradientValues.data)
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -1211,11 +1211,11 @@ cdef class Interface:
         self.thisptr.readScalarData (data_id, vertex_id, _value)
         return _value
 
-def write_block_vector_gradient_data (self, data_id, vertex_ids, gradientValues):
+        def write_block_vector_gradient_data (self, data_id, vertex_ids, gradientValues):
         """
-        Writes vector data given as block. This function writes values of specified vertices to a dataID.
+        Writes vector gradient data given as block. This function writes values of specified vertices to a dataID.
         Values are provided as a block of continuous memory. Values are stored in a numpy array [N x D] where N = number
-        of vertices and D = dimensions of geometry
+        of vertices and D = dimensions of geometry squared.
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,223 @@ cdef class Interface:
         self.thisptr.readScalarData (data_id, vertex_id, _value)
         return _value
 
+    def write_block_vector_gradient_data (self, data_id, vertex_ids, gradientValues):
+        """
+        Writes vector gradient data given as block. This function writes values of specified vertices to a dataID.
+        Values are provided as a block of continuous memory. Values are stored in a numpy array [N x D] where N = number
```
DavidSCN, line 1217: I'm not sure if the memory comment here is valuable for python users. I guess you copied this over from the c++ docs?

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,223 @@ cdef class Interface:
...
+        of vertices and D = dimensions of geometry squared.
```
DavidSCN, line 1218:
```suggestion
        of vertices and D = number of gradient components.
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,223 @@ cdef class Interface:
...
+
+        Parameters
+        ----------
+        data_id : int
+            Data ID to write to.
+        vertex_ids : array_like
+            Indices of the vertices.
+        gradientValues : array_like
+             Gradient values differentiated in the spacial direction (dx, dy) for 2D space, (dx, dy, dz) for 3D space
```
DavidSCN, line 1227: From what you write above, the `gradientValues` array should have 'number of gradient components' many entries. Thus, there is missing information about the format, right?

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,223 @@ cdef class Interface:
...
+
+        Notes
+        -----
+        Previous calls:
+            Count of available elements at values matches the configured dimension
```
DavidSCN, line 1232: Is this a copy-left artifact? Which values are you referring to? Do you mean gradients here?

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,223 @@ cdef class Interface:
...
+        Write block gradient vector data for a 2D problem with 2 vertices:
+        >>> data_id = 1
+        >>> vertex_ids = [1, 2]
+        >>> gradientValues = np.array([[v1x_dx, v1y_dx,v1x_dy, v1y_dy], [v2x_dx, v2y_dx,v2x_dy, v2y_dy]])
+        >>> interface.write_block_vector_data(data_id, vertex_ids, gradientValues)
+
+        Write block vector data for a 3D (D=3) problem with 2 (N=2) vertices:
+        >>> data_id = 1
+        >>> vertex_ids = [1, 2]
+        >>> values = np.array([[v1x_dx, v1y_dx, v1z_dx, v1x_dy, v1y_dy, v1z_dy, v1x_dz, v1y_dz, v1z_dz], [v2x_dx, v2y_dx, v2z_dx, v2x_dy, v2y_dy, v2z_dy, v2x_dz, v2y_dz, v2z_dz]])
```
DavidSCN, line 1248:
```suggestion
        >>> gradientValues = np.array([[v1x_dx, v1y_dx, v1z_dx, v1x_dy, v1y_dy, v1z_dy, v1x_dz, v1y_dz, v1z_dz], [v2x_dx, v2y_dx, v2z_dx, v2x_dy, v2y_dy, v2z_dy, v2x_dz, v2y_dz, v2z_dz]])
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,223 @@ cdef class Interface:
...
+            Count of available elements at vertex_ids matches the given size
+            Initialize() has been called
+            Data with dataID has attribute hasGradient = true
+            
+        Examples
+        --------
+        Write block gradient vector data for a 2D problem with 2 vertices:
+        >>> data_id = 1
+        >>> vertex_ids = [1, 2]
+        >>> gradientValues = np.array([[v1x_dx, v1y_dx,v1x_dy, v1y_dy], [v2x_dx, v2y_dx,v2x_dy, v2y_dy]])
```
DavidSCN, line 1242: Looks like a nested array instead of an `[N x D]` array. Maybe I'm missing something?

DavidSCN (commented):
Looks good overall, the documentation needs some clarification. Please apply relevant comments to all API functions (didn't look at all). The docker image has been updated. As far as I understand you can add tests now and they should work.

cyprecice/cyprecice.pyx
kursatyurt: It is from python bindings, https://github.com/kursatyurt/python-bindings/blob/bd7d0315c7db73b996a7879944c05ef645ca5a5c/cyprecice/cyprecice.pyx#L872-L875

cyprecice/cyprecice.pyx
kursatyurt: Format is shown in examples, writing both scalar and vector format seems to just replicate the examples below https://github.com/kursatyurt/python-bindings/blob/bd7d0315c7db73b996a7879944c05ef645ca5a5c/cyprecice/cyprecice.pyx#L1237-L1250

cyprecice/cyprecice.pyx
kursatyurt: Yes it should be `Count of available elements at values matches the number of gradient components * size` right?

cyprecice/cyprecice.pyx
kursatyurt: It is an (2,4) array N=2 D=4 (2 Dimension, 4 Gradient Value)

```python
import numpy as np
a = np.array([[1, 2,3, 4], [5, 6,7, 8]])
a.shape
Out[4]: (2, 4)
```

cyprecice/cyprecice.pyx
kursatyurt: Thank you. I missed.

kursatyurt:
@IshaanDesai @BenjaminRodenberg @DavidSCN  Tests are ready. Since gradient data has no read API, tests are only write-type tests that checks different data structures.

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,249 @@ cdef class Interface:
...
+
+        Returns
+        -------
+        bool
+            True if gradient data is required for a dataID.
+
+        Notes
+        -----
+        Previous calls:
+            initialize() has been called
```
DavidSCN, line 1446: That's not a requirement from preCICE. Is it one for python?

cyprecice/cyprecice.pyx
DavidSCN: In case you write this, I still don't know which values you are referring to. There are no values mentioned anywhere.

DavidSCN:
> Since gradient data has no read API, tests are only write-type tests that checks different data structures.

How can we validate the correctness of the data passed to preCICE? Can we execute a basic integration test or access preCICE internals?

IshaanDesai:
> How can we validate the correctness of the data passed to preCICE? Can we execute a basic integration test or access preCICE internals?

I have not looked at the tests yet but in general the idea of mock testing in the python bindings is that we provide an expected output to a mocked preCICE interface and then we pass the data and compare it against the expected (correct) data already present inside the mocked preCICE interface.

cyprecice/cyprecice.pyx
IshaanDesai: I suspect this is just a copy paste from other functions. @kursatyurt please make sure that the required previous calls are really required for the new API.

kursatyurt: @DavidSCN I took it from, https://github.com/precice/precice/blob/d83d3c789172acca414233bd65524297c4ce3438/src/precice/SolverInterface.hpp#L1142 

Can we really call it before the initialization of preCICE?

DavidSCN: In doxygen, the documentation appears above the function declaration. In python, the documentation appears below the function declaration (right?). The `isRequired` function does not require `initialize`. The `writeGradient` functions require `initialize`.

cyprecice/cyprecice.pyx
kursatyurt: I think the same here in C++ doc
https://github.com/precice/precice/blob/d83d3c789172acca414233bd65524297c4ce3438/src/precice/SolverInterface.hpp#L1165 
 
 https://github.com/precice/python-bindings/blob/ed5de3dbb6fd29faf1421c2c009db3bf1d5cf203/cyprecice/cyprecice.pyx#L1217-L1218
 
Here it mentioned `[NxD]` so I think `Count of available elements at values matches the configured dimension` was enough or more explicitly we can write `Count of available elements at values matches the number of gradient components`

kursatyurt:
> > How can we validate the correctness of the data passed to preCICE? Can we execute a basic integration test or access preCICE internals?
> 
> I have not looked at the tests yet but in general the idea of mock testing in the python bindings is that we provide an expected output to a mocked preCICE interface and then we pass the data and compare it against the expected (correct) data already present inside the mocked preCICE interface.

So can I write the assertion in `cpp` file and expect some correct input from the test script?

cyprecice/cyprecice.pyx
kursatyurt: All clear, thank you.

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
         self.thisptr.readScalarData (data_id, vertex_id, _value)
         return _value
 
+    def write_block_vector_gradient_data (self, data_id, vertex_ids, gradientValues):
+        """
+        Writes vector gradient data given as block. This function writes values of specified vertices to a dataID.
```
IshaanDesai, line 1216:
```suggestion
        Writes vector gradient data given as block. This function gradient writes values of specified vertices to a dataID.
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
+            Initialize() has been called
+            Data with dataID has attribute hasGradient = true
+            
+        Examples
+        --------
+        Write block gradient vector data for a 2D problem with 2 vertices:
+        >>> data_id = 1
+        >>> vertex_ids = [1, 2]
+        >>> gradientValues = np.array([[v1x_dx, v1y_dx,v1x_dy, v1y_dy], [v2x_dx, v2y_dx,v2x_dy, v2y_dy]])
+        >>> interface.write_block_vector_data(data_id, vertex_ids, gradientValues)
```
IshaanDesai, line 1243:
```suggestion
        >>> interface.write_block_vector_gradient__data(data_id, vertex_ids, gradientValues)
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
+
+        Write block vector data for a 3D (D=3) problem with 2 (N=2) vertices:
+        >>> data_id = 1
+        >>> vertex_ids = [1, 2]
+        >>> gradientValues = np.array([[v1x_dx, v1y_dx, v1z_dx, v1x_dy, v1y_dy, v1z_dy, v1x_dz, v1y_dz, v1z_dz], [v2x_dx, v2y_dx, v2z_dx, v2x_dy, v2y_dy, v2z_dy, v2x_dz, v2y_dz, v2z_dz]])
+        >>> interface.write_block_vector_data(data_id, vertex_ids, gradientValues)
```
IshaanDesai, line 1249:
```suggestion
        >>> interface.write_block_vector_gradient_data(data_id, vertex_ids, gradientValues)
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
+            Count of available elements at vertex_ids matches the given size
+            Initialize() has been called
+            Data with dataID has attribute hasGradient = true
+            
+        Examples
+        --------
+        Write block gradient vector data for a 2D problem with 2 vertices:
+        >>> data_id = 1
+        >>> vertex_ids = [1, 2]
+        >>> gradientValues = np.array([[v1x_dx, v1y_dx,v1x_dy, v1y_dy], [v2x_dx, v2y_dx,v2x_dy, v2y_dy]])
```
IshaanDesai, line 1242:
```suggestion
        >>> gradientValues = np.array([[v1x_dx, v1y_dx, v1x_dy, v1y_dy], [v2x_dx, v2y_dx, v2x_dy, v2y_dy]])
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
+        """
+        check_array_like(vertex_ids, "vertex_ids", "write_block_vector_gradient_data")
+        check_array_like(gradientValues, "gradientValues", "write_block_vector_gradient_data")
+
+        if not isinstance(gradientValues, np.ndarray):
+            gradientValues = np.asarray(gradientValues)
+
+        if len(gradientValues) > 0:
+            size, dimensions = gradientValues.shape
+            assert dimensions == self.get_dimensions() * self.get_dimensions(), "Dimensions of vector data in write_block_vector_gradient_data does not match with dimensions in problem definition. Provided dimensions: {}, expected dimensions: {}".format(dimensions, self.get_dimensions() *  self.get_dimensions())
```
IshaanDesai, line 1259: Is this assertion really correct? From the example you have written above, the dimensions you get from ´
`gradientValues.shape` would still be 2 or 3 and 4 or 9

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
+        if len(gradientValues) == 0:
+            size = 0
+
+        cdef np.ndarray[int, ndim=1] _vertex_ids = np.ascontiguousarray(vertex_ids, dtype=np.int32)
+        cdef np.ndarray[double, ndim=1] _gradientValues = np.ascontiguousarray(gradientValues.flatten(), dtype=np.double)
+
+        assert _gradientValues.size == size * self.get_dimensions() * self.get_dimensions(), "Vector data is not provided for all vertices in write_block_vector_data. Check length of input data provided. Provided size: {}, expected size: {}".format(_gradientValues.size, size * self.get_dimensions() * self.get_dimensions())
```
IshaanDesai, line 1266:
```suggestion
        assert _gradientValues.size == size * self.get_dimensions() * self.get_dimensions(), "Vector data is not provided for all vertices in write_block_vector_gradient_data. Check length of input data provided. Provided size: {}, expected size: {}".format(_gradientValues.size, size * self.get_dimensions() * self.get_dimensions())
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
+
+        Previous calls:
+            initialize() has been called
+
+        Examples
+        --------
+        Write scalar data for a 2D problem with 5 vertices:
+        >>> data_id = 1
+        >>> vertex_id = 5
+        >>> gradientValue = [v5_dx,v5_dy]
```
IshaanDesai, line 1307:
```suggestion
        >>> gradientValue = [v5_dx, v5_dy]
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
+        Count of available elements at value matches the configured dimension
+        Vertex with dataID exists and contains data
+        Data with dataID has attribute hasGradient = true
+
+        Previous calls:
+            initialize() has been called
+
+        Examples
+        --------
+        Write scalar data for a 2D problem with 5 vertices:
```
IshaanDesai, line 1304:
```suggestion
        Write scalar data for a 2D problem:
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
```
IshaanDesai, line 1355:
```suggestion
        Write scalar data for a 2D problem:
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
+        """
+
+        check_array_like(gradientValues, "gradientValues", "write_vector_gradient_data")
+
+        if not isinstance(gradientValues, np.ndarray):
+            gradientValues = np.asarray(gradientValues)
+
+        cdef np.ndarray[double, ndim=1] _gradientValues = np.ascontiguousarray(gradientValues.flatten(), dtype=np.double)
+
+        assert _gradientValues.size == self.get_dimensions()* self.get_dimensions(), "Vector data provided for vertex {} in write_scalar_gradient_data does not match problem definition. Check length of input data provided. Provided size: {}, expected size: {}".format(_gradientValues.size, self.get_dimensions() * self.get_dimensions())
```
IshaanDesai, line 1369:
```suggestion
        assert _gradientValues.size == self.get_dimensions() * self.get_dimensions(), "Vector data provided for vertex {} in write_vector_gradient_data does not match problem definition. Check length of input data provided. Provided size: {}, expected size: {}".format(_gradientValues.size, self.get_dimensions() * self.get_dimensions())
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
+            Count of available elements at vertex_ids matches the given size
+            Initialize() has been called
+            Data with dataID has attribute hasGradient = true
+            
+        Examples
+        --------
+        Write block gradient vector data for a 2D problem with 2 vertices:
+        >>> data_id = 1
+        >>> vertex_ids = [1, 2]
+        >>> gradientValues = np.array([[v1_dx, v1_dy],[v2_dx, v2_dy]])
```
IshaanDesai, line 1401:
```suggestion
        >>> gradientValues = np.array([[v1_dx, v1_dy], [v2_dx, v2_dy]])
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
+        -----
+        Previous calls:
+            Count of available elements at values matches the configured dimension
+            Count of available elements at vertex_ids matches the given size
+            Initialize() has been called
+            Data with dataID has attribute hasGradient = true
+            
+        Examples
+        --------
+        Write block gradient vector data for a 2D problem with 2 vertices:
```
IshaanDesai, line 1398:
```suggestion
        Write block gradient scalar data for a 2D problem with 2 vertices:
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
+        >>> data_id = 1
+        >>> vertex_ids = [1, 2]
+        >>> gradientValues = np.array([[v1_dx, v1_dy],[v2_dx, v2_dy]])
+        >>> interface.write_block_vector_data(data_id, vertex_ids, gradientValues)
```
IshaanDesai, line 1402:
```suggestion
        >>> interface.write_block_scalar_gradient_data(data_id, vertex_ids, gradientValues)
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
+
+        Write block vector data for a 3D (D=3) problem with 2 (N=2) vertices:
+        >>> data_id = 1
+        >>> vertex_ids = [1, 2]
+        >>> values = np.array([[v1_dx, v1_dy, v1x_dz], [v2_dx. v2_dy, v2_dz]])
+        >>> interface.write_block_scalar_data(data_id, vertex_ids, values)
```
IshaanDesai, line 1408:
```suggestion
        Write block scalar data for a 3D (D=3) problem with 2 (N=2) vertices:
        >>> data_id = 1
        >>> vertex_ids = [1, 2]
        >>> values = np.array([[v1_dx, v1_dy, v1x_dz], [v2_dx. v2_dy, v2_dz]])
        >>> interface.write_block_scalar_gradient_data(data_id, vertex_ids, values)
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
+        """
+        check_array_like(vertex_ids, "vertex_ids", "write_block_scalar_gradient_data")
+        check_array_like(gradientValues, "gradientValues", "write_block_sclar_gradient_data")
+
+        if not isinstance(gradientValues, np.ndarray):
+            gradientValues = np.asarray(gradientValues)
+
+        if len(gradientValues) > 0:
+            size, dimensions = gradientValues.shape
+            assert dimensions == self.get_dimensions() , "Dimensions of vector data in write_block_scalar_gradient_data does not match with dimensions in problem definition. Provided dimensions: {}, expected dimensions: {}".format(dimensions, self.get_dimensions())
```
IshaanDesai, line 1418:
```suggestion
            assert dimensions == self.get_dimensions() , "Dimensions of scalar gradient data  provided in write_block_scalar_gradient_data does not match with dimensions in problem definition. Provided dimensions: {}, expected dimensions: {}".format(dimensions, self.get_dimensions())
```

cyprecice/cyprecice.pyx
```diff
@@ -1211,6 +1211,244 @@ cdef class Interface:
...
+        if len(gradientValues) == 0:
+            size = 0
+
+        cdef np.ndarray[int, ndim=1] _vertex_ids = np.ascontiguousarray(vertex_ids, dtype=np.int32)
+        cdef np.ndarray[double, ndim=1] _gradientValues = np.ascontiguousarray(gradientValues.flatten(), dtype=np.double)
+
+        assert _gradientValues.size == size * self.get_dimensions(), "Vector data is not provided for all vertices in write_block_vector_data. Check length of input data provided. Provided size: {}, expected size: {}".format(_gradientValues.size, size * self.get_dimensions())
```
IshaanDesai, line 1425:
```suggestion
        assert _gradientValues.size == size * self.get_dimensions(), "Scalar gradient data is not provided for all vertices in write_block_scalar_gradient_data. Check length of input data provided. Provided size: {}, expected size: {}".format(_gradientValues.size, size * self.get_dimensions())
```

cyprecice/cyprecice.pyx
DavidSCN: Sounds a bit weird

test/test_bindings_module.py
```diff
@@ -409,3 +409,107 @@ def test_get_mesh_vertices_and_ids(self):
...
+    def test_is_gradient_data_required(self):
+        solver_interface = precice.Interface("test", "dummy.xml", 0, 1)
+        fake_bool = 0  # compare to output in test/SolverInterface.cpp
+        fake_data_id = 0
+        self.assertEqual(fake_bool, solver_interface.is_gradient_data_required(fake_data_id))
+
+    def test_write_block_scalar_gradient_data(self):
+        solver_interface = precice.Interface("test", "dummy.xml", 0, 1)
+        write_data = np.array([[1, 2, 3], [6, 7, 8], [9, 10, 11]])
+        solver_interface.write_block_scalar_gradient_data(1, np.array([1, 2, 3]), write_data)
```
IshaanDesai, line 422: This test is not really testing the bindings API, and I understand that the problem is because gradient data functions are only write functions and we dont have read functions to get data back and check if the correct data was passed on to the preCICE API. Lets do the following, write the `write_data` to the mocked preCICE interface and then access the `fake_gradient_buffer` inside the mocked interface to check if the data is equal to `write_data`. The check can be done with `self.assertTrue(np.array_equal(write_data, solver_interface.fake_gradient_buffer))`.

**Note: This applies to all the tests below**

cyprecice/cyprecice.pyx
IshaanDesai: > Sounds a bit weird

Better now? A word was in the wrong place

DavidSCN: Awesome

test/test_bindings_module.py
kursatyurt: It seems we need another solution since on the cpp side `fake_gradient_buffer` is just a static variable 
```    self.assertTrue(np.array_equal(write_data, solver_interface.fake_gradient_buffer))
AttributeError: 'cyprecice.Interface' object has no attribute 'fake_gradient_buffer'

```

IshaanDesai: Okay I understand the problem. Thinking more on this I have concluded that the only way to test such write-only functionality is to add something to the Python bindings API itself. Lets add a private function to the Python bindings API which returns the gradient data written. This function will be used only for testing and will be hidden from the user. @kursatyurt can you find a nice way to do this?

DavidSCN: Free function sounds like a good idea. Which errors do you face? Can you post them here? If you define (and use) the free function only in the `cpp` file I would consider wrapping it into an anonymous namespace.

IshaanDesai: The idea of having a free function in the local header file sounds really good! That way we can use it for testing and do not have to add anything in the bindings API. Lets debug the errors you are facing :+1:

IshaanDesai: I have a simpler idea to handle the problem we are facing. Why dont we just write the gradient data to the `fake_read_write_buffer` and then return is using the read commands? That way we can check if the gradient data is correct inside the API and dont have to implement additional read functionality anywhere.

kursatyurt: @IshaanDesai block versions would work, other way around it gets a bit more complex, I will change to this. Thank you very much

IshaanDesai: > block versions would work, other way around it gets a bit more complex

What do you mean by the *other way around* here? I would just use the corresponding read functionality, so if you are testing `write_block_scalar_gradient_data`, I would use `read_block_scalar_data` to get the data back, and same for all other other API functions.

kursatyurt: > > block versions would work, other way around it gets a bit more complex
> 
> What do you mean by the _other way around_ here? I would just use the corresponding read functionality, so if you are testing `write_block_scalar_gradient_data`, I would use `read_block_scalar_data` to get the data back, and same for all other API functions.

Writing new functions to return gradient data and keep them updated. 

I have used `read_block_scalar_data` for all of them. The problem here is dimensions of return arrays do not match since the gradient of the scalar is a vector, etc. 

It is a bit ugly, but I have to convert all the input to a 1D nd.array for comparison.

Now mostly works. I have problem with `test_write_block_scalar_gradient_data_single_float`. I ended up with a segmentation fault. Could you check that?

IshaanDesai: > Now mostly works. I have problem with `test_write_block_scalar_gradient_data_single_float`. I ended up with a segmentation fault. Could you check that?

Fixed :+1:  I'll clean up the tests a bit before merging this

IshaanDesai (approved):
API and tests look good now :+1: I will try to update the Spack image so that the build goes through and then we can merge this

davidscn:
@IshaanDesai is this now ready to merge?

IshaanDesai:
@davidscn @kursatyurt just sync with develop and then we can proceed with merging

kursatyurt:
> @davidscn @kursatyurt just sync with develop and then we can proceed with merging

Done!.

PR #149, opened 2022-08-22
Repairing warnings and error regarding Version.h by using pkgconfig to link to preCICE

IshaanDesai:
This PR resolves a warning and an error in the action set of `build-and-test`. The warning was:

```bash
warning: cyprecice/cyprecice.pyx:1211:57: local variable '_value' referenced before assignment
```
and the error was
```bash
/home/runner/work/python-bindings/python-bindings/precice/SolverInterface.hpp:7:10: fatal error: precice/Version.h: No such file or directory
    7 | #include "precice/Version.h"
      |          ^~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -45,7 +45,7 @@ jobs:
           rm -rf /var/lib/apt/lists/*
           pip3 install --upgrade --user pip
       - name: Check preCICE linkage via pkg-config
-        run: pkg-config --modversion precice
+        run: pkg-config --modversion libprecice
       - name: Install dependencies
         run: |
           pip3 install --user toml
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -97,7 +97,7 @@ jobs:
           cd precice-core
           mkdir build && cd build
           cmake .. -DPRECICE_MPICommunication=OFF -DPRECICE_PETScMapping=OFF -DPRECICE_PythonActions=OFF -DBUILD_TESTING=OFF
-          cp src/precice/Version.h ~/precice/
+          cp src/precice/Version.h ../../precice/
       - name: Install dependencies
         run: |
           python3 -c 'import toml; c = toml.load("pyproject.toml"); print("\n".join(c["build-system"]["requires"]))' | pip3 install -r /dev/stdin
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -34,7 +34,7 @@ jobs:
   setup_install:
     name: Run setup install
     runs-on: ubuntu-latest
-    container: precice/precice:2.4.0
+    container: precice/precice
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -84,7 +84,7 @@ jobs:
       - name: Install OpenMPI
         run: |
           sudo apt-get -yy update
-          sudo apt-get install -y libopenmpi-dev cmake
+          sudo apt-get install -y libopenmpi-dev cmake libboost-all-dev
           sudo rm -rf /var/lib/apt/lists/*
       - uses: BSFishy/pip-action@v1
         with:
@@ -96,7 +96,7 @@ jobs:
           cp precice-core/src/precice/SolverInterface.hpp precice/SolverInterface.hpp
           cd precice-core
           mkdir build && cd build
-          cmake ..
+          cmake .. -DPRECICE_MPICommunication=OFF -DPRECICE_PETScMapping=OFF -DPRECICE_PythonActions=OFF -DBUILD_TESTING=OFF
           cp src/precice/Version.h ~/precice/
       - name: Install dependencies
         run: |
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -101,6 +101,7 @@ jobs:
           PKG_CONFIG_PATH: "precice-core/build"
       - name: Install dependencies
         run: |
+          echo "$PKG_CONFIG_PATH"
           python3 -c 'import toml; c = toml.load("pyproject.toml"); print("\n".join(c["build-system"]["requires"]))' | pip3 install -r /dev/stdin
       - name: Check preCICE linkage via pkg-config
         run: pkg-config --modversion libprecice
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -12,7 +12,7 @@ jobs:
     name: Run setup.py phases needed by spack
     needs: [setup_test]
     runs-on: ubuntu-latest
-    container: precice/precice
+    container: precice/precice:2.4.0
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
@@ -34,7 +34,7 @@ jobs:
   setup_install:
     name: Run setup install
     runs-on: ubuntu-latest
-    container: precice/precice
+    container: precice/precice:2.4.0
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
@@ -59,7 +59,7 @@ jobs:
     name: Run setup install --single-version-externally-managed (for spack)
     needs: [setup_install]
     runs-on: ubuntu-latest
-    container: precice/precice
+    container: precice/precice:2.4.0
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
@@ -113,7 +113,7 @@ jobs:
     name: Run pip install
     needs: [setup_test]
     runs-on: ubuntu-latest
-    container: precice/precice
+    container: precice/precice:2.4.0
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -94,6 +94,10 @@ jobs:
           git clone https://github.com/precice/precice.git precice-core
           mkdir -p precice
           cp precice-core/src/precice/SolverInterface.hpp precice/SolverInterface.hpp
+          cd precice-core
+          mkdir build && cd build
+          cmake .. -DPRECICE_MPICommunication=OFF -DPRECICE_PETScMapping=OFF -DPRECICE_PythonActions=OFF -DBUILD_TESTING=OFF
+          cp src/precice/Version.h ../../precice/
```
fsimonis, line 100: This will break the moment we add another generated header.

A long-term solution would be to install preCICE and use it from that location.
A even-longer-term solution is to use pkg-config in the setup.py to detect preCICE.

fsimonis (changes requested):

.github/workflows/build-and-test.yml
IshaanDesai: Yes it is true that this would break immediately when we add another generated header, but for the mock testing of the python-bindings I need to not have preCICE built on the system, so I am left with the option to either create the header via cmake and copy it over, or delete the include line from `SolverInterface.hpp`, but the later solution will also break with another generated header. Is there a solution for when preCICE is not built?

fsimonis: You could use pkg-config for the include-arguments only. This works from the build directory after executing CMake.
Then you can build and provide linker flags to your mocked library.

.github/workflows/build-and-test.yml
```diff
@@ -94,6 +94,11 @@ jobs:
           git clone https://github.com/precice/precice.git precice-core
           mkdir -p precice
           cp precice-core/src/precice/SolverInterface.hpp precice/SolverInterface.hpp
+          cd precice-core
+          mkdir build && cd build
+          cmake .. -DPRECICE_MPICommunication=OFF -DPRECICE_PETScMapping=OFF -DPRECICE_PythonActions=OFF -DBUILD_TESTING=OFF
+        env:
+          PKG_CONFIG_PATH: "precice-core/build/lib/pkgconfig"
```
fsimonis, line 101: The one you need is in the build directory. The other one is for the installation.
```suggestion
          PKG_CONFIG_PATH: "precice-core/build"
```

IshaanDesai:
In finding a solution to include the generated header `Version.h` in the mock testing workflow, the Python package `pkgconfig` is used to link the bindings to preCICE. This implementation was already discussed and partially done in https://github.com/precice/python-bindings/pull/135, and a full solution in consultation with @fsimonis is done here.

.github/workflows/build-and-test.yml
```diff
@@ -12,14 +12,14 @@ jobs:
     name: Run setup.py phases needed by spack
     needs: [setup_test]
     runs-on: ubuntu-latest
-    container: precice/precice
```
BenjaminRodenberg, line 15: I'm not a big fan of this change. Why is it necessary? Be aware that you will need to update this for every single release of preCICE. Additionally: You won't be testing against the develop version of preCICE, but against the last released version. This leaves you with a blind spot between the latest release and the develop version and might lead to bad surprises when you update.

However, I guess you cannot do a lot here. Ideally there would be a `precice/precice:develop` pointing to the develop version of preCICE and a `precice/precice:latest` pointing to the latest release. Not sure which one you would need here.

.github/workflows/build-and-test.yml
```diff
@@ -34,14 +34,14 @@ jobs:
   setup_install:
     name: Run setup install
     runs-on: ubuntu-latest
-    container: precice/precice
```
BenjaminRodenberg, line 37: See above.

.github/workflows/build-and-test.yml
```diff
@@ -57,14 +57,14 @@ jobs:
     name: Run setup install --single-version-externally-managed (for spack)
     needs: [setup_install]
     runs-on: ubuntu-latest
-    container: precice/precice
+    container: precice/precice:2.4.0
```
BenjaminRodenberg, line 60: See above.

.github/workflows/build-and-test.yml
```diff
@@ -106,14 +112,14 @@ jobs:
     name: Run pip install
     needs: [setup_test]
     runs-on: ubuntu-latest
-    container: precice/precice
+    container: precice/precice:2.4.0
```
BenjaminRodenberg, line 115: See above

.github/workflows/run-solverdummy.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Install Dependencies
         run: |
           apt-get -qq update
-          apt-get -qq install software-properties-common python3-dev python3-pip git apt-utils
+          apt-get -qq install software-properties-common python3-dev python3-pip git apt-utils pkg-config
```
BenjaminRodenberg, line 20: Why don't you use consequently `precice/precice:2.4.0` here as well? I'm a `precice/precice` fan, but I'm still wondering why you are making the difference here.

BenjaminRodenberg (commented):
Overall I think using pkgconfig makes a lot of sense here.

The issue with `precice/precice` vs. `precice/precice:2.4.0` might be an upstream issue, but maybe this is also an argument to change things upstream?

One thing that might be affected by this change: Do you need updates for the spack recipe?

.github/workflows/build-and-test.yml
IshaanDesai: Good catch! This was leftover from when I was testing things on my local machine using `act`. Indeed one does not need a version tag here!

fsimonis:
> One thing that might be affected by this change: Do you need updates for the spack recipe?

Yes.
Currently, spack is failing due to `py-pkgconfig` missing during build. I assume this based on the build failing during processing of the `pyproject.toml` file.

.github/workflows/build-and-test.yml
```diff
@@ -94,10 +94,16 @@ jobs:
           git clone https://github.com/precice/precice.git precice-core
           mkdir -p precice
           cp precice-core/src/precice/SolverInterface.hpp precice/SolverInterface.hpp
+          cd precice-core
+          mkdir build && cd build
+          cmake .. -DPRECICE_MPICommunication=OFF -DPRECICE_PETScMapping=OFF -DPRECICE_PythonActions=OFF -DBUILD_TESTING=OFF
```
BenjaminRodenberg, line 99: Do you really need this? The purpose of the mocking was that we don't have to install preCICE. The line above where we copy `precice-core/src/precice/SolverInterface.hpp` should be sufficient. Can you try removing these 3 lines again? 

But if it is needed for pkg-config to work properly I guess it is just as it is. As long as the mock implementation from `python-bindings/test/SolverInterface.cpp` is used properly it is fine.

BenjaminRodenberg: Actually, pkg-config should not be used here. You are running the tests and the tests don't use pkg-config for linking to preCICE, but the hard-coded path with the mock-version of the interface.

.github/workflows/build-and-test.yml
```diff
@@ -84,7 +84,7 @@ jobs:
       - name: Install OpenMPI
         run: |
           sudo apt-get -yy update
-          sudo apt-get install -y libopenmpi-dev
+          sudo apt-get install -y libopenmpi-dev cmake libboost-all-dev libeigen3-dev pkg-config
```
BenjaminRodenberg, line 87: If you don't have to install preCICE (see below), you should be able to get away without these dependencies.

.github/workflows/build-and-test.yml
```diff
@@ -84,7 +84,7 @@ jobs:
       - name: Install OpenMPI
```
BenjaminRodenberg, line 84: If you install all preCICE dependencies, this should be renamed.

BenjaminRodenberg (changes requested):
I still have some minor points w.r.t to the mocking of preCICE. Can you try to remove the build process of preCICE and the installation of the dependencies, if it is not absolutely needed? I think it should work without building preCICE.

.github/workflows/build-and-test.yml
IshaanDesai: For the mock tests we need to run cmake because the `SolverInterface.hpp` that we copy over, has an included header `Version.h`. This header file is a generated header file, so we need to run cmake to get the file. After the file is generated, for the mock tests we use the cflags from pkg-config to link to this generated header. So we actually do use pkg-config even for the mock testing, just not the library flags.

.github/workflows/build-and-test.yml
IshaanDesai: I am not installing preCICE completely, just running the cmake command, hence I need these dependencies.

.github/workflows/build-and-test.yml
BenjaminRodenberg: Ah ok. So I guess this is what this whole PR was about. Makes sense :+1:

.github/workflows/build-and-test.yml
BenjaminRodenberg: Understood :+1:

BenjaminRodenberg (approved):
After some clarifications I cannot find anything to complain anymore. Thanks for the explanations @IshaanDesai :pray:

PR #159, opened 2023-01-10
Make bindings compatible with precice:develop

IshaanDesai:
This PR makes the develop branch of the bindings compatible with precice:develop.

### DIFF
--- a/cyprecice/SolverInterface.pxd
+++ b/cyprecice/SolverInterface.pxd
@@ -33,12 +33,6 @@ cdef extern from "precice/SolverInterface.hpp" namespace "precice":
 
         bool hasToEvaluateFineModel () const
 
-        # action methods
-
-        bool isActionRequired (const string& action) const
-
-        void markActionFulfilled (const string& action)
-
         # mesh access
 
         bool hasMesh (const string& meshName ) const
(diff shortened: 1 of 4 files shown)


Timo-Schrader:
Sounds good, I don't have any other important changes currently.

BenjaminRodenberg:
Installation (`pip3 install --user .`) and tests  (`python3 setup.py test`) work locally on my system if I use [8c52300](https://github.com/precice/python-bindings/pull/159/commits/8c523001d9fbd554152bc27da689880e5df066b1). I think there are some problems at the moment (that we cannot really fix here):

* The Dockerimage `precice/precice` used by the tests is not pointing to develop, but to an older version with the outdated API. There is some work going on in this direction in https://github.com/precice/setup-precice-action/pull/1 and https://github.com/precice/precice/issues/1544 in this direction.
* I have problems with the installation procedure of preCICE when using the current version of develop. See https://github.com/precice/precice/issues/1554 for details.

IshaanDesai:
There are no unpushed changes from my side too. This is a pending todo in my list, and I am happy if you push this @BenjaminRodenberg  :)

BenjaminRodenberg:
As far as I understand, the main problem why we cannot merge this PR is that the tests don't work with the new action `precice-setup-action`. Therefore, I removed this part of the testing procedure and instead use my own docker image `benjaminrodenberg/precice:develop`. See #169 

We can still keep this PR and merge it as soon as the `precice-setup-action` is working. This would then upgrade the tests to this new work flow.

PR #162, opened 2023-01-11
Adapted compatibility-precice-dev branch

Timo-Schrader:
- Added docstrings for some functions
- Removed `return` statement in l. 593 of `cyprecice/cyprecice.pxy` (please confirm that this is correct)
- Removed `import` statements in `__init__.py` since these functions do no longer exist.


BenjaminRodenberg (approved):
All looks good to me :+1: I did not check it thoroughly, but let's merge this one quick and then we continue working in https://github.com/precice/python-bindings/pull/159 together. Thank you!

PR #164, opened 2023-02-13
set PKG_CONFIG_PATH if necessary

nkr0:
When installed in a non-system location, pkg-config is not able to find the configuration file. According to pkg-config's manual, "the default is libdir/pkgconfig:datadir/pkgconfig where libdir is the libdir where pkg-config and datadir is the datadir **where pkg-config was installed.**"

This PR sets `PKG_CONFIG_PATH` using `LD_LIBRARY_PATH` to make finding precice easier.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,13 +61,11 @@ APPNAME = "pyprecice"
 PYTHON_BINDINGS_PATH = os.path.dirname(os.path.abspath(__file__))
 
 
-def set_pkg_config_libdir():
-    if pkgconfig.exists("libprecice"):
-        return
-    for path in os.environ["LD_LIBRARY_PATH"].split(":"):
-        os.environ["PKG_CONFIG_LIBDIR"] = path + "/pkgconfig"
-        if pkgconfig.exists("libprecice"):
-            return
+def set_pkg_config_path():
+    if not pkgconfig.exists("libprecice"):
+        os.environ["PKG_CONFIG_PATH"] = ":".join(
+            [path + "/pkgconfig" for path in os.environ["LD_LIBRARY_PATH"].split(":")]
+        )
 
 
 def get_extensions(is_test):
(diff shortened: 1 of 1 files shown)


nkr0:
For when precice is installed in non system places. I'll add a few comments in the file for explanation.

setup.py
```diff
@@ -61,6 +61,13 @@
 PYTHON_BINDINGS_PATH = os.path.dirname(os.path.abspath(__file__))
 
 
+def set_pkg_config_path():
+    if not pkgconfig.exists("libprecice"):
```
nkr0, line 65: pkconfig documentation says that it'll search for libs **where pkg-config was installed**. This will be system directories if pkg-config was installed from distro repos (so most cases). If precice was compiled by the user and installed in a non-system location, `pkgconfig.exists("libprecice")` will be `False` with the default settings.

setup.py
```diff
@@ -61,6 +61,13 @@
...
+        os.environ["PKG_CONFIG_PATH"] = ":".join(
+            [path + "/pkgconfig" for path in os.environ["LD_LIBRARY_PATH"].split(":")]
```
nkr0, line 67: If we then assume that the user has configured `LD_LIBRARY_PATH` for dynamic linking of `libprecice` (because it is in a non-system location), we can then use that as `PKG_CONFIG_PATH` to find the library.

IshaanDesai:
@nkr0 thanks for the explanation, now its clear what the change does. I think this makes sense, and if I remember correctly, you opened this PR after a discussion with @fsimonis on Gitter, right?

IshaanDesai (approved):

fsimonis (commented):
The proposed solution is only necessary in environments that are not fully setup. We fully cover this in our documentation.

Furthermore, we are actively trying to migrate as much logic as possible from the `setup.py` to `pyproject.toml`. #208 

I suggest closing this and instead focus on improving the error message pkg-config can't find preCICE by linking to the documentation.

IshaanDesai:
Closing this in favor of handling the error message when preCICE has not bee found, in https://github.com/precice/python-bindings/issues/209

PR #166, opened 2023-02-16
Provide Waveform API for v2

BenjaminRodenberg:
This PR adds the Waveform API added in preCICE v2.4.0 to the python bindings.


.github/workflows/build-and-test.yml
```diff
@@ -91,7 +91,7 @@ jobs:
           packages: toml
       - name: Checkout precice and make required files discoverable
         run: |
-          git clone https://github.com/precice/precice.git precice-core
+          git clone -b v2.5.0 https://github.com/precice/precice.git precice-core
```
BenjaminRodenberg, line 94: We need this, because `develop` is already at `v3.x` state.

BenjaminRodenberg:
I think this PR is fine. All the errors I get in the tests here also appear on `develop`, so I assume we don't fix them here.

PR #168, opened 2023-02-17
Release v2.5.0.2

IshaanDesai:
Main feature:

* Add Waveform API

### DIFF
--- a/cyprecice/SolverInterface.pxd
+++ b/cyprecice/SolverInterface.pxd
@@ -95,19 +95,19 @@ cdef extern from "precice/SolverInterface.hpp" namespace "precice":
 
         void readBlockVectorData (const int dataID, const int size, const int* valueIndices, double* values) const
 
-        void readBlockVectorData (const int dataID, const int size, const int* valueIndices, double dt, double* values) const
+        void readBlockVectorData (const int dataID, const int size, const int* valueIndices, double relativeReadTime, double* values) const
 
         void readVectorData (const int dataID, const int valueIndex, double* value) const
 
-        void readVectorData (const int dataID, const int valueIndex, double dt, double* value) const
+        void readVectorData (const int dataID, const int valueIndex, double relativeReadTime, double* value) const
 
         void readBlockScalarData (const int dataID, const int size, const int* valueIndices, double* values) const
 
-        void readBlockScalarData (const int dataID, const int size, const int* valueIndices, double dt, double* values) const
+        void readBlockScalarData (const int dataID, const int size, const int* valueIndices, double relativeReadTime, double* values) const
 
         void readScalarData (const int dataID, const int valueIndex, double& value) const
 
-        void readScalarData (const int dataID, const int valueIndex, double dt, double& value) const
+        void readScalarData (const int dataID, const int valueIndex, double relativeReadTime, double& value) const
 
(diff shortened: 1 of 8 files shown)


BenjaminRodenberg (approved):
Good from my perspective, but I am also the one responsible for the most controversial changes :smiling_imp:

cyprecice/SolverInterface.pxd
```diff
@@ -95,13 +95,21 @@ cdef extern from "precice/SolverInterface.hpp" namespace "precice":
 
         void readBlockVectorData (const int dataID, const int size, const int* valueIndices, double* values) const
 
+        void readBlockVectorData (const int dataID, const int size, const int* valueIndices, double dt, double* values) const
```
fsimonis, line 98: Is there any reason why the relative read time is called dt in this function? 
It may be a good idea to keep the name consistent across and inside bindings.

cyprecice/cyprecice.pyx
```diff
@@ -1090,10 +1092,13 @@ cdef class Interface:
         size = _vertex_ids.size
         dimensions = self.get_dimensions()
         cdef np.ndarray[np.double_t, ndim=1] _values = np.empty(size * dimensions, dtype=np.double)
-        self.thisptr.readBlockVectorData (data_id, size, <const int*>_vertex_ids.data, <double*>_values.data)
+        if relative_read_time == None:
```
fsimonis, line 1095: Makes a small difference :wink:

```suggestion
        if relative_read_time is None:
```

cyprecice/cyprecice.pyx
```diff
@@ -1104,6 +1109,8 @@ cdef class Interface:
             ID to read from.
         vertex_id : int
             Index of the vertex.
+        relative_read_time : double
+            Point in time where data is read relative to the beginning of the current time step
```
fsimonis, line 1113: What should a user do if they want a relative time of 0?
Pass 0, pass None, or omit the argument?

BenjaminRodenberg: Pass `relative_read_time=0`. This points to the beginning of the current time *step*. The API is designed such that you always get the value at the end of the window, if you pass nothing (or `None`), because this is the current behavior of preCICE. This is equivalent to calling a read function with `relative_read_time = dt`, where `dt = precice.advance(...)`. Does this help?

fsimonis: That helps me, but not the reader of the documentation :wink:

IshaanDesai: I will finalize this for now, let us revisit the documentation if there is indeed confusion on the user side. We want to release this API for a first user, so we shall know then.

fsimonis (approved):

cyprecice/cyprecice.pyx
BenjaminRodenberg: @IshaanDesai Documentation of the CPP API is much more detailed. See [here](https://github.com/precice/precice/blob/fcefb9d6620f8d4f8dfefa574bbcd44002ce7adf/src/precice/SolverInterface.hpp#L880-L910). I guess keeping this synchronized is indeed a different issue. Ideally this would happen automatically, but I know... Let's not worry about this here.

PR #169, opened 2023-02-17
Make bindings compatible with precice:develop

### LINKED ISSUE #155: Remove API function get_mesh_vertex_ids_from_positions and get_mesh_vertices
Deprecate the bindings API function `get_mesh_vertex_ids_from_positions` because the original API function `getMeshVertexIDsFromPositions` is being deprecated: https://github.com/precice/precice/issues/1248.

### LINKED ISSUE #161: Undefined Symbol Error when using preCICE Python bindings
**Describe your setup**

Operating system (e.g. Linux distribution and version): Debian 11 Bullseye
preCICE Version: Latest (develop)

preCICE Python bindings: precice/python-bindings@71e7a6e38af3418192b3d8f65006bc8f97a1d4dd

**Describe the problem**

I am currently fixing the preCICE Python bindings s.t. they are compatible with 632b4fbfa5608b57eb1b46e897cb8a049d09847b and ec11b6e918151b22d5d6d78a3da4d03c3b3d059b. This requires building the Python bindings from scratch and linking it against preCICE dynamically. However, this produces the following error messages during import in Python even though libprecice.so is accessible:

`ImportError: ~/.local/lib/python3.9/site-packages/cyprecice.cpython-39-x86_64-linux-gnu.so: undefined symbol: _ZN7precice9cplscheme9constants22actionWriteInitialDataB5cxx11Ev`

Looking into the ELF symbol table, it gets clear that `actionWriteInitialData` has a _LOCAL_ binding instead of _GLOBAL_ which makes it inaccessible for the linker. I found the following setting to be the root cause of the problem, which has been introduced a month ago:

https://github.com/precice/precice/blob/dd70dc238a87d9263370d09dc320779c2d72dcbb/CMakeLists.txt#L264

Therefore, we either need to change this setting s.t. the functions `actionWriteInitialData`, `actionWriteIterationCheckpoint`, `actionReadIterationCheckpoint` and so on are accessible again in the shared library of preCICE or we need to remove them from the Python bindings.

**Step To Reproduce**
1. Build preCICE Python bindings from scratch.
2. Run `import precice` in Python.

**Expected behaviour**
preCICE gets imported without any linking errors. Removing the visibility flag from CMakeLists.txt fixes that.

BenjaminRodenberg:
This PR makes the develop branch of the bindings compatible with precice:develop.

Closes #161, closes #155

This PR contains all the changes from #159, but uses a simpler testing procedure.

**important** The docker image `benjaminrodenberg/precice:develop` is not maintained. Meaning: We will have to update it manually. This is currently a workaround, because I think cannot wait until https://github.com/precice/precice/issues/1115 or https://github.com/precice/precice/issues/1544 is resolved.


BenjaminRodenberg:
@IshaanDesai can you review this PR? I think we can ignore spack. All other tests are working now. Feel free to directly squash & merge, if you think that everything is ok.

IshaanDesai (approved):
Looks good :+1: I did not check every line, but the passing tests are already enough of a check.

PR #174, opened 2023-03-09
Use mesh name and data name instead of IDs

IshaanDesai:
API changes are in line with https://github.com/precice/precice/pull/1588

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -86,10 +86,8 @@ jobs:
       - uses: actions/setup-python@v1
       - name: Install OpenMPI, CMake, Boost library, Eigen and pkg-config
         run: |
-          su precice
           apt-get -yy update
           apt-get install -y libopenmpi-dev cmake libboost-all-dev libeigen3-dev pkg-config
-          rm -rf /var/lib/apt/lists/*
       - uses: BSFishy/pip-action@v1
         with:
           packages: toml
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -82,7 +82,6 @@ jobs:
         uses: actions/checkout@v2
       - uses: actions/setup-python@v1
       - name: Install OpenMPI, CMake, Boost library, Eigen and pkg-config
-        user: root
         run: |
           apt-get -yy update
           apt-get install -y libopenmpi-dev cmake libboost-all-dev libeigen3-dev pkg-config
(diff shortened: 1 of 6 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -82,11 +86,10 @@ jobs:
...
       - uses: actions/setup-python@v1
       - name: Install OpenMPI, CMake, Boost library, Eigen and pkg-config
-        user: root
         run: |
-          apt-get -yy update
-          apt-get install -y libopenmpi-dev cmake libboost-all-dev libeigen3-dev pkg-config
-          rm -rf /var/lib/apt/lists/*
+          sudo apt-get -yy update
+          sudo apt-get install -y libopenmpi-dev cmake libboost-all-dev libeigen3-dev pkg-config
+          sudo rm -rf /var/lib/apt/lists/*
```
BenjaminRodenberg, line 92: Why `sudo` here and `su root` above?

.github/workflows/build-and-test.yml
```diff
@@ -95,6 +98,8 @@ jobs:
           git clone --branch develop https://github.com/precice/precice.git precice-core
           mkdir -p precice
           cp precice-core/src/precice/SolverInterface.hpp precice/SolverInterface.hpp
+          cp precice-core/src/precice/Tooling.hpp precice/Tooling.hpp
+          cp precice-core/src/precice/Tooling.hpp precice/Tooling.cpp
```
BenjaminRodenberg, line 102: Does this belong into this PR? Is this not more a general compatibility thing for the testing?

BenjaminRodenberg (commented):
Hard to judge, because there are many changes. But if the tests work, then it should be fine.

Main thing that confuses me: Do the updates in the CI pipeline really belong here?

.github/workflows/build-and-test.yml
IshaanDesai: This change is indeed not part of the main change in this pull request, but has been included here to get the tests to work. The tests are currently breaking on develop anyway, so pushing changes separately leads the tests to break on practically every feature branch.

.github/workflows/build-and-test.yml
IshaanDesai: Removed the `su root` above as one is already root in the `precice:precice/develop` container. Here we need `root` because we are in a general `ubuntu:latest` container.

.github/workflows/build-and-test.yml
```diff
@@ -18,6 +18,7 @@ jobs:
         uses: actions/checkout@v2
       - name: Install & upgrade pip3
         run: |
+          su root
```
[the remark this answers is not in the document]
IshaanDesai: Good catch! I removed it.

.github/workflows/build-and-test.yml
IshaanDesai: Thanks for the suggestions, its done now :+1:

fsimonis (approved):
Looks like a sane change. Let's get this merged so that we can test these design changes in the adapters.

Issue #176, opened 2023-05-09
pip install --user pyprecice - fails Cython dependency


mennodeij:
If I issue the command to locally get the pyprecice installation, it fails. 
After I first install Cython, it works. It would be nice if the installation of pyprecice automatically first installed Cython.

```
pip3 install --user pyprecice
Collecting pyprecice
  Using cached https://files.pythonhosted.org/packages/c2/85/868d99a904a7e08ebb2adba932e481ff0614d6d25d1ff0c848f3b777f4c2/pyprecice-2.5.0.2.tar.gz
    Complete output from command python setup.py egg_info:
    /tmp/pip-build-4ks8hq0e/pyprecice/setup.py:33: UserWarning: It looks like you are trying to use pip for installation of the package. Please install, the module packaging by running 'pip3 install --user packaging', since it is needed to perform additional security checks. You can continue installation. However, if you face problems when installing or running pyprecice, it might be a good idea to install packaging to enable additional checks.
      "It looks like you are trying to use pip for installation of the package. Please install, "
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-4ks8hq0e/pyprecice/setup.py", line 51, in <module>
        from Cython.Distutils.extension import Extension
    ModuleNotFoundError: No module named 'Cython'

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-4ks8hq0e/pyprecice/
```

IshaanDesai:
I remember that there was a discussion with @BenjaminRodenberg at some point on why we do not explicitly give cython as a dependency required at the time of installation. But I no longer remember the reasoning and it makes sense to have a look at this.

BenjaminRodenberg:
@mennodeij how did you install cython in the end to make it work? I assume via pip, but better to be sure.

@IshaanDesai Unfortunately I don't remember the reasoning for not providing cython and currently I also don't see a reason. Maybe this had something to do with spack, but I thing we can try out to add the dependency on a branch and see whether our tests fail. For some of the tricky cases with spack we have tests so it should be relatively safe.

IshaanDesai:
> Maybe this had something to do with spack, but I thing we can try out to add the dependency on a branch and see whether our tests fail. For some of the tricky cases with spack we have tests so it should be relatively safe.

I will create a branch and test this.

mennodeij:
Yes, it was installed via `pip install Cython --user`

PR #177, opened 2023-05-10
Add Cython as dependency required during installation

IshaanDesai:
Possibly resolves https://github.com/precice/python-bindings/issues/176


BenjaminRodenberg (approved):
I cannot see a (documented) reason speaking against adding it and if it is causing trouble without, we obviously need it.

PR #179, opened 2023-05-30
Compatibility with precice:develop

IshaanDesai:
This PR makes the develop branch of the bindings compatible with the latest state of precice:develop.


cyprecice/cyprecice.pyx
```diff
@@ -242,41 +280,44 @@ cdef class Interface:
         """
         return self.thisptr.requiresInitialData ()
 
-    def requires_reading_checkpoint (self):
+
+    def requires_writing_checkpoint (self):
```
BenjaminRodenberg, line 284: I guess you swapped the order of these two functions to make it consistent with `precice/src/precice/Participant.hpp`? Makes the diff a bit messy, but makes sense. Maybe consider moving this minor change as an individual commit directly to develop to improve readability here.

cyprecice/cyprecice.pyx
```diff
@@ -699,186 +776,55 @@ cdef class Interface:
...
 
-        cdef np.ndarray[int, ndim=1] _vertex_ids = np.ascontiguousarray(vertex_ids, dtype=np.int32)
-        cdef np.ndarray[double, ndim=1] _values = np.ascontiguousarray(values, dtype=np.double)
+        assert len(vertex_ids) == size, "Vertex IDs are of incorrect length in write_data. Check length of vertex ids input. Provided size: {}, expected size: {}".format(vertex_ids.size, size)
 
-        assert _values.size == size, "Scalar data is not provided for all vertices in write_block_scalar_data. Check size of input data provided. Provided size: {}, expected size: {}".format(_values.size, size)
-        assert _vertex_ids.size == size, "Vertex IDs are of incorrect length in write_block_scalar_data. Check size of vertex ids input. Provided size: {}, expected size: {}".format(_vertex_ids.size, size)
+        cdef vector[int] cpp_ids = vertex_ids
+        cdef vector[double] cpp_values = values.flatten()
```
BenjaminRodenberg, line 820: Would be good, if we could avoid this explicit copying of data and use ranges instead. But let's do this as a follow-up.

cyprecice/cyprecice.pyx
```diff
@@ -484,29 +523,27 @@ cdef class Interface:
...
 
-        cdef bytes mesh_name_py_bytes = mesh_name.encode()
+        self.thisptr.setMeshEdges (convert(mesh_name), cpp_vertices)
 
-        self.thisptr.setMeshEdges (<const char*> mesh_name_py_bytes, size, <const int*>_vertices.data)
 
     def set_mesh_triangle (self, mesh_name, first_vertex_id, second_vertex_id, third_vertex_id):
         """
-        Sets mesh triangle from edge IDs
+        Set a mesh triangle from edge IDs
```
BenjaminRodenberg, line 546: Can you try to move these minor documentation updates outside of this PR to improve readability of the diff? Not critical, but this PR is already hard enough to read.

examples/solverdummy/solverdummy.py
```diff
@@ -20,56 +20,60 @@
 participant_name = args.participantName
 
 if participant_name == 'SolverOne':
-    write_data_name = 'dataOne'
-    read_data_name = 'dataTwo'
-    mesh_name = 'MeshOne'
+    write_data_name = 'Data-One'
+    read_data_name = 'Data-Two'
+    mesh_name = 'SolverOne-Mesh'
```
BenjaminRodenberg, line 25: Maybe move this into another commit/PR? Again: Not 100% related to this update and makes this PR bigger than necessary.

BenjaminRodenberg (approved):
There are a lot of changes in this PR and this makes the diff hard to read. One reason is clearly the renaming of some important files and classes and I think we cannot really improve readability here. Still, it would be nice to extract some minor parts into independent commits for a nicer git history.

Besides readability: From the feature-side everything looks ok (tests work on my system)

PR #180, opened 2023-06-13
Lint markdown files and update files correspondingly.

BenjaminRodenberg:


BenjaminRodenberg:
I decided to ignore some rules, because they would require some additional updates that are not straight-forward. Documentation needs an update at some point in time anyways. But for now this is sufficient.

PR #186, opened 2023-08-02
Adapt Docker workflow

valentin-seitz:
This PR adopts the current Dockerfile to match 
- the precice convention of the tag `latest` beeing used for the latest stable release -> master branch
- a more rigid naming schema of Build Arguments needed for the Systemtests. 
This is also needed in the downstream components like the fenics-adapter (https://github.com/precice/fenics-adapter/pull/161)

It also introduces a manual triggering of the rebuilt process. 

**Please note:**  the `push` workflow trigger mechanism could not be tested.


.github/workflows/build-docker.yml
```diff
@@ -13,10 +22,26 @@ jobs:
     env:
         docker_username: precice
     steps:
-      - name: Get branch name
-        if: github.event_name != 'pull_request'
+      - name: Set branch name for manual triggering
+        if: github.event_name == 'workflow_dispatch'
+        shell: bash
+        run: echo "BINDINGS_REF=${{ inputs.branch }}" >> $GITHUB_ENV
+      - name: Set branch name for on pull triggering
```
MakisH, line 29:
```suggestion
      - name: Set branch name for "on pull request" triggering
```

.github/workflows/build-docker.yml
```diff
@@ -31,6 +56,7 @@ jobs:
         with:
           push: true
           file: "./tools/releasing/packaging/docker/Dockerfile"
-          tags: ${{ env.docker_username }}/python-bindings:${{ env.branch }},${{ env.docker_username }}/python-bindings:latest
+          tags: ${{ env.docker_username }}/python-bindings:${{ env.TAG }}
```
MakisH, line 59: This is the main change / fix.
Now, if the branch is `master`, the tag applied is `latest`. Otherwise, it is the name of the branch / Git reference.

tools/releasing/packaging/docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Dockerfile to build a ubuntu image containing the installed Debian package of a release 
-ARG branch=develop
-ARG from=precice/precice:latest
+ARG PRECICE_TAG=develop
+ARG from=precice/precice:${PRECICE_TAG}
```
MakisH, line 3: It looks like currently the CI is always using the `latest` tag of the preCICE Docker image, which is currently (correctly) v2.5.0.

This is an update we anyway should have, but it may already trigger some error in the CI.

.github/workflows/build-docker.yml
```diff
@@ -13,10 +22,26 @@ jobs:
...
         docker_username: precice
     steps:
-      - name: Get branch name
-        if: github.event_name != 'pull_request'
+      - name: Set branch name for manual triggering
+        if: github.event_name == 'workflow_dispatch'
+        shell: bash
+        run: echo "BINDINGS_REF=${{ inputs.branch }}" >> $GITHUB_ENV
+      - name: Set branch name for on pull triggering
+        if: github.event_name != 'pull_request' && github.event_name != 'workflow_dispatch'
```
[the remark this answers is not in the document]
valentin-seitz: This is the "automatic mode" when a push on either of the branches `master` or `develop` happens. (bascially after a PR)

tools/releasing/packaging/docker/Dockerfile
valentin-seitz: I mean i already pushed a `precice/python-bindings:latest` image yesterday and for me the fenics adapter built nicely with it.

IshaanDesai (changes requested):
Looks good, but why are we merging this to `master` directly?

MakisH (approved):
All requested changes implemented, switched target branch to develop.

MakisH:
Looking at the [published image on Docker Hub](https://hub.docker.com/r/precice/python-bindings/tags), the images with tags `develop` and `latest` are currently the same (same digest).

After being merged to master, this PR will change the `latest` to point to the state of `master`. A workaround till then is to [trigger the workflow manually](https://github.com/precice/python-bindings/actions/workflows/build-docker.yml). After merging, this workflow gives the option for the branch to use.

PR #192, opened 2024-02-06
Update release guide

IshaanDesai:
During the release of v3.0.0.0, I noticed that some instructions in the release guide are not easy to follow and a bit confusing. I reordered some things and also moved the pre-releasing points into their own section.

For reviewing, it would make sense if one directly reads the new release guide and tries to gauge the flow of the document.


docs/ReleaseGuide.md
```diff
@@ -13,21 +13,27 @@ The release of the `python-bindings` repository is made directly from a release
...
+4. [Draft a New Release](https://github.com/precice/python-bindings/releases/new) in the `Releases` section of the repository page in a web browser.
 
-    * *Note:* We use the [python-versioneer](https://github.com/python-versioneer/python-versioneer/) for maintaining the version. Therefore the tag directly defines the version in all relevant places.
-    * *Note:* If it is a pre-release then the option *This is a pre-release* needs to be selected at the bottom of the page. Use `@target:python-bindings-v2.1.1.1` for a pre-release, since we will never merge a pre-release into master.
-    * Use the `Auto-generate release notes` feature. See, for example, https://github.com/precice/python-bindings/releases/tag/v2.3.0.1
+    * The release tag needs to be the exact version number (i.e.`v2.1.1.1` or `v2.1.1.1rc1`, compare to [existing tags](https://github.com/precice/python-bindings/tags)).
+    * If this is a full release, use `@target:master`. If this is a pre-release, use `@target:python-bindings-v2.1.1.1`. If you are making a pre-release, **directly skip to the [pre-release](#pre-release) section below**.
```
BenjaminRodenberg, line 19:
```suggestion
    * If this is a stable release, use `@target:master`. If this is a pre-release, use `@target:python-bindings-v2.1.1.1`. If you are making a pre-release, **directly skip to the [pre-release](#pre-release) section below**.
```
Better wording? https://en.wikipedia.org/wiki/Software_release_life_cycle

docs/ReleaseGuide.md
```diff
@@ -13,21 +13,27 @@ The release of the `python-bindings` repository is made directly from a release
...
 
-    a) If a pre-release is made: Directly hit the "Publish release" button in your Release Draft. Now you can check the artifacts (e.g. release on [PyPI](https://pypi.org/project/pyprecice/#history)) of the release. *Note:* As soon as a new tag is created github actions will take care of deploying the new version on PyPI using [this workflow](https://github.com/precice/python-bindings/actions?query=workflow%3A%22Upload+Python+Package%22).
+5. As soon as one approving review is made, merge the release PR (from `python-bindings-v2.1.1.1`) into `master`.
 
-    b) If this is a "real" release: As soon as one approving review is made, merge the release PR (from `python-bindings-v2.1.1.1`) into `master`.
+6. Merge `master` into `develop` for synchronization of `develop`.
 
-5. Merge `master` into `develop` for synchronization of `develop`.
+7. If everything is in order up to this point then the new version can be released by hitting the "Publish release" button in your Release Draft. This will create the corresponding tag and trigger [publishing the release to PyPI](https://github.com/precice/python-bindings/actions?query=workflow%3A%22Upload+Python+Package%22).
```
BenjaminRodenberg, line 26:
```suggestion
7. If everything is in order up to this point then the new version can be released by hitting the "Publish release" button in your release Draft. This will create the corresponding tag and trigger [publishing the release to PyPI](https://github.com/precice/python-bindings/actions?query=workflow%3A%22Upload+Python+Package%22).
```

docs/ReleaseGuide.md
```diff
@@ -13,21 +13,27 @@ The release of the `python-bindings` repository is made directly from a release
...
 
-6. If everything is in order up to this point then the new version can be released by hitting the "Publish release" button in your Release Draft. This will create the corresponding tag and trigger [publishing the release to PyPI](https://github.com/precice/python-bindings/actions?query=workflow%3A%22Upload+Python+Package%22).
+8. Add an empty commit on master by running the steps:
```
BenjaminRodenberg, line 28:
```suggestion
8. Add an empty commit (details https://github.com/precice/python-bindings/issues/109) on master by running the steps:
```

docs/ReleaseGuide.md
```diff
@@ -48,4 +54,8 @@ The release of the `python-bindings` repository is made directly from a release
 
    For more details refer to https://github.com/precice/python-bindings/issues/109 and https://github.com/python-versioneer/python-versioneer/issues/217.
 
-8. Update Spack package (refer to `python-bindings/spack/README.md`).
+9. *Temporarily not maintained* Update Spack package (refer to `python-bindings/spack/README.md`).
+
+## Pre-release
+
+A pre-release branch is never merged into a master branch. After creating the branch and drafting a release, directly hit the "Publish release" button in your Release Draft. Now you can check the pre-release artifacts (e.g. release on [PyPI](https://pypi.org/project/pyprecice/#history)) of the release. No further action is required for a pre-release.
```
BenjaminRodenberg, line 61:
```suggestion
The purpose of a pre-release is to test the release workflow (packaging, upload to PyPI). After drafting a release, directly hit the "Publish release" button in your Release Draft. Now you can check the pre-release artifacts (e.g. release on [PyPI](https://pypi.org/project/pyprecice/#history)) of the release. No further action is required for a pre-release.
```

Is this actually true? Aren't we just testing the release workflow with a pre-release? We can actually create a pre-release from the release branch that will later be merged into master. This makes perfectly sense.

BenjaminRodenberg (changes requested):
I think the pre-release workflow is still a bit strange: We should do the pre-release from the same branch and commit that should be used as a basis for the stable release in the future. Remember: We want to test the workflow of the actual release and should , therefore, stay as close to the actual release as possible.

I think some tweaking of 4. is necessary.

docs/ReleaseGuide.md
IshaanDesai: Creating a pre-release from the release branch itself makes sense, we should do this for the future. But until now we never did this.

IshaanDesai:
@BenjaminRodenberg I reworded the pre-release. Can you have a look once more?

BenjaminRodenberg:
Needs merge, but looks good to me :+1:

BenjaminRodenberg (approved):

Issue #191, opened 2024-02-06
Docker image workflow fails on the master branch


IshaanDesai:
Due to changes in https://github.com/precice/python-bindings/blame/develop/.github/workflows/build-docker.yml#L36-L44 the [workflow](https://github.com/precice/python-bindings/actions/workflows/build-docker.yml) to update the Docker image is triggered on the master branch. This workflow fails on the master branch: https://github.com/precice/python-bindings/actions/runs/7799939617/job/21271746402 because it tries to build the Docker image using the `precice/precice:latest` image, which does not have user `precice`. Previously this did not happen because the workflow was only run for the develop branch.

 The question now is, should we change something in the Docker recipe of the bindings or of preCICE to address the error, or shall we just build the Docker image of the bindings always on `precice/precice:develop`.

MakisH:
## What the actual error is

Triggered from a push to the `master` branch, the [`build-and-release-docker-image` workflow](https://github.com/precice/python-bindings/blob/c0288c880e64a20848d2275c2b057e5041ee057e/.github/workflows/build-docker.yml#L23) is attempting to install the bindings as a user (`precice`), but the user does not exist:

```
ERROR: failed to solve: process "/bin/bash -c python3 -m pip install --user --upgrade pip" did not complete successfully: unable to find user precice: no matching entries in passwd file
```

## Why this is triggered now

The changes in #186 only enabled the workflow to use the correct preCICE Docker image tag when building on `master` and made a few more things configurable when running the workflow manually. They also enabled executing the workflow on `master`, as we (may) need such a Docker image to exist for some use cases of the system tests.

So, while this only appears now, it was always an issue that was not triggered. The changes of @valentin-seitz are correct, provided that both images are consistent (keep reading).

It was not triggered so far, because it would only be triggered on `master` when the workflow changes would be merged into `master`.

## What triggers this error

Indeed, the problem here is the current inconsistency in the setup between the [`precice/precice:latest`](https://hub.docker.com/layers/precice/precice/latest/images/sha256-00939883168544811cdb2618f9370597a4f6d31aaa9400657d17a9ff3677e005?context=explore) and the [`precice/precice:develop`](https://hub.docker.com/layers/precice/precice/develop/images/sha256-f8b9877477ccf2865012ca9de70d0c6487262ef517646d7ccf981b4f24a3b81f?context=explore) images on Dockerhub.

These originate from this GitHub Actions workflow of preCICE:

[...]

PR #196, opened 2024-02-13
Disable build-docker workflow on push to master

MakisH:
This disables building the Docker image on push to master, which was first enabled in #186. This workflow was simply not triggered before, but currently fails #191.

Publishing such a Docker image is not needed for the system tests anymore.

@BenjaminRodenberg @IshaanDesai I am directly contributing to the master branch as a hotfix, since this only fixes (disables) the broken GitHub Actions workflow.

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -8,7 +8,6 @@ on:
         description: branch to build the image from
         options:
           - develop
-          - master     
   push:
     branches:
       - develop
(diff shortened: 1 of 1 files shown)


MakisH:
This of course only "fixes" the GitHub view. I can also fully revert all other changes to the Dockerfile etc and revert to the previous state (which would still break if anyone tried to build for master, but would never see that).

MakisH:
Note also that, as I am not a maintainer of this repository, I have no idea what side-implications pushing to `master` has, so proceed with caution. I claim no responsibility for such side effects.

BenjaminRodenberg:
I think the main question here is what we actually want. I already discussed this to some degree with @fsimonis.

1. If we want to release docker images with the python bindings - this also implies docker images for downstream projects like the fenics adapter - I don't think this is a good solution.
2. If we only need docker images for testing, then we can probably skip the workflow for master and only regularly build a new version of `precice/python-bindings:develop`

My personal opinion: I can currently live without having an official docker release of the python bindings and the fenics adapter. Having the develop images is fine. With the releases on PyPI and conda I think our users also have good means to get the packages.

IshaanDesai:
> My personal opinion: I can currently live without having an official docker release of the python bindings and the fenics adapter. Having the develop images is fine. With the releases on PyPI and conda I think our users also have good means to get the packages.

:white_check_mark:

BenjaminRodenberg (approved):

BenjaminRodenberg:
@MakisH I removed some code that should now have become unnecessary. Can you do a final check and then merge?

MakisH:
Looks good to me. You merge. :-)

PR #204, opened 2024-07-04
Fix numpy include order

fsimonis:
This PR fixes the numpy include order.

This leads to the system numpy being considered after local installations (venv or pip --user).

This is the reason why cython failed to compile when using numpy>=2 with a system numpy<2.


IshaanDesai (approved):
Looks good :+1: this will require a bugfix release so that packages using the Python bindings do not see the Cython compilation failure.

PR #205, opened 2024-07-08
Release v3.1.1

IshaanDesai:
This is a bugfix release of the bindings. The order in which NumPy is included in the build setup is changed to prevent a conflict between the system installed NumPy, and the NumPy installed via pip.


fsimonis (approved):

BenjaminRodenberg (approved):
Oscillator works fine :+1:

PR #207, opened 2024-07-09
Require setuptools>=61

BenjaminRodenberg:
Requiring setuptools>=61 allows us to remove a noisy warning that always shows up when building with spack. It also allows us to simplify the `setup.py`. Setuptools>=61 guarantees that pyproject.toml is used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file.
 
+## latest
+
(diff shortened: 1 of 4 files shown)


fsimonis (approved):
Good way forward. Now pyproject.toml :smile:

BenjaminRodenberg:
> Good way forward. Now pyproject.toml 😄

I've put this into #208

PR #212, opened 2024-08-27
Release v2.5.1.0

BenjaminRodenberg:
Backports some trivial changes to v2.5.0.
Compatibility release for preCICE [v2.5.1](https://github.com/precice/precice/releases/tag/v2.5.1).
Closes #203

PR #213, opened 2024-08-27
Restrict to numpy < 2 and setuptools < 72 for better compatibiltiy with CI pipeline

BenjaminRodenberg:
I would suggest to restrict the numpy version to `numpy<2` since the newest numpy is not well supported by our CI pipeline (see #212).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## latest
 
 * Restrict to numpy < 2 for better compatibility with CI pipeline. https://github.com/precice/python-bindings/pull/213
+* Require setuptools < 72 since support for the test command was removed in Setuptools 72. https://github.com/precice/python-bindings/pull/213
 * Require setuptools >= 61 to guarantee that pyproject.toml is used https://github.com/precice/python-bindings/pull/207
 * Fix CI pipeline for spack https://github.com/precice/python-bindings/pull/206
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## latest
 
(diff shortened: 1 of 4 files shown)


IshaanDesai (approved):
Looks reasonable :+1:

BenjaminRodenberg:
Will merge this now. One review should be sufficient.

fsimonis:
@BenjaminRodenberg I do understand these restrictions for the v2.5.1 release to keep it simple, but not for develop.

numpy 2 works fine after #204. The bindings now actively prevent downstream projects (adapters and cases) from updating to numpy 2. This is a problem.

The setuptools restriction only affects testing. The rest works fine.
If we want to test, then we could instead install a working setuptools in the CI until we have a fix.
If there are special version requirements of the spack package should be handled in the spack package.

BenjaminRodenberg:
> @BenjaminRodenberg I do understand these restrictions for the v2.5.1 release to keep it simple, but not for develop.
> 
> numpy 2 works fine after #204. The bindings now actively prevent downstream projects (adapters and cases) from updating to numpy 2. This is a problem.
> 
> The setuptools restriction only affects testing. The rest works fine. If we want to test, then we could instead install a working setuptools in the CI until we have a fix. If there are special version requirements of the spack package should be handled in the spack package.

Thanks for the info. I was not completely aware of this connection. We then basically need the restrictions as introduced here only for pyprecice 2.x; maybe for some pyprecice 3.x releases prior to https://github.com/precice/python-bindings/pull/204. For all pyprecice 3.x releases after https://github.com/precice/python-bindings/pull/204 we should loosen the restriction on numpy for the reasons you already mentioned above.

I'll open an issue for this that we don't forget. With spack it's unfortunately always a bit tricky to get things working properly since I'm not that deep into spack and properly testing different configurations takes some time.

PR #218, opened 2024-09-04
Allow numpy 2 for pyprecice>=3.1.1

BenjaminRodenberg:
Closes https://github.com/precice/python-bindings/issues/219

To some degree I would also consider this as a bugfix since Numpy support was implemented for `3.1.1` in #204 and broken in `3.1.2` via #213.


BenjaminRodenberg:
No need to worry about our spack CI too much since we will remove it soon. See https://github.com/precice/python-bindings/pull/221. Will merge this PR now.

PR #221, opened 2024-09-04
Remove spack testing

### LINKED ISSUE #220: Spack testing obsolete
The tests of the python bindings to keep spack working involves many layers as spack used to install packages by calling setup.py as follows:

```
$ python setup.py build
$ python setup.py install --root=...
```

As the python bindings are native extensions of precice using cython, this involved additional steps that needed testing.

Spack changed to directly using `pip install` in https://github.com/spack/spack/pull/27798.
Meaning that we no longer need to test `setup.py` calls and the installation of the python bindings becomes significantly simpler and more robust.

I suggest to:
* remove testing of setup.py calls which are anyhow deprecated.
* remove testing of spack builds in the CI, including the generation of the build image and the checked-in spack package files.
* remove the explicit checks for a pip environment in the `setup.py`. 

This allows us to more aggressively push #208 and #217.

fsimonis:
This PR removes the spack-related tests and documentation.

Testing the spack package is now an additional optional step in the Release Guide.

Closes #220


.github/workflows/build-and-test.yml
```diff
@@ -8,120 +8,8 @@ on:
...
-      - name: Install dependencies
-        run: |
-          pip3 install --user toml
-          python3 -c 'import toml; c = toml.load("pyproject.toml"); print("\n".join(c["build-system"]["requires"]))' | pip3 install -r /dev/stdin
-      - name: Run setup install
-        run:  python3 setup.py install --user
-      - name: Test install
-        run: |
-          export LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
-          python3 -c "import precice"
```
BenjaminRodenberg, line 60: The standard installation using `python3 setup.py install --user` we should still keep (maybe update to `pip install .`?

.github/workflows/build-and-test.yml
```diff
@@ -8,120 +8,8 @@ on:
...
-      - name: Install dependencies
-        run: |
-          python3 -c 'import toml; c = toml.load("pyproject.toml"); print("\n".join(c["build-system"]["requires"]))' | pip3 install -r /dev/stdin
-      - name: Run setup test
-        env:
-          PKG_CONFIG_PATH: "precice-core/build"
-          PKG_CONFIG_SYSTEM_INCLUDE_PATH: 1
-        run: |
-          export CFLAGS=-I$GITHUB_WORKSPACE
-          python3 setup.py test
```
BenjaminRodenberg, line 120: These are the mock unit tests. We should keep them.

.github/workflows/build-and-test.yml
```diff
@@ -8,120 +8,8 @@ on:
...
-
   pip_install:
     name: Run pip install
-    needs: [setup_test]
```
BenjaminRodenberg, line 124: Not 100% sure what this does. Can you explain why you remove it?

.github/workflows/build-and-test.yml
```diff
@@ -144,7 +32,7 @@ jobs:
 
   solverdummy_test:
     name: Run solverdummy
-    needs: [setup_install, setup_test]
+    needs: [pip_install]
```
BenjaminRodenberg, line 35: Similar to above.

.github/workflows/build-env.yml
BenjaminRodenberg, line 1: Good to remove this :+1:

.github/workflows/build-spack.yml
BenjaminRodenberg, line 1: Why not regularly try to build the latest develop version of the bindings using spack? (potentially with the develop version of package.py)

README.md
```diff
@@ -59,66 +59,12 @@ If preCICE (the C++ library) was installed in a custom prefix, or only built but
...
-   ```bash
-   $ python3 setup.py install --user
-   ```
-
-5. Clean-up *optional*
-
-   ```bash
-   $ python3 setup.py clean --all
-   ```
-
```
BenjaminRodenberg, line 121: I think this does not belong into this PR.

docs/ReleaseGuide.md
```diff
@@ -13,21 +13,23 @@ The release of the `python-bindings` repository is made directly from a release
     * `CHANGELOG.md` on `python-bindings-v2.1.1.1`.
     * There is no need to bump the version anywhere else, since we use the [python-versioneer](https://github.com/python-versioneer/python-versioneer/) for maintaining the version everywhere else.
 
-4. [Draft a New Release](https://github.com/precice/python-bindings/releases/new) in the `Releases` section of the repository page in a web browser.
+4. *Optional* test the [py-pyprecice Spack package](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/py-pyprecice/package.py) using `spack dev-build py-pyprecice@develop`.
```
BenjaminRodenberg, line 16: Sounds good to introduce this as a manual step :+1: 

Do we really want to do this as optional? Additionally: This might take quite some time. Maybe this is an argument to just regularly do this step in the CI?

spack/repo/packages/py-pyprecice/package.py
BenjaminRodenberg, line 1: Do we really not need any `package.py`? We could, of course, also just maintain the `package.py` in a fork of the spack repo and completely ignore spack here. But then I think we need some more detailed instructions in the `ReleaseGuide.md`

BenjaminRodenberg (changes requested):
Cleaning this up makes a lot of sense. But I'm still not sure how the new process for updating the spack package will look like.

I think you also accidentally deleted some tests that we will still need.

.github/workflows/build-and-test.yml
fsimonis: Invoking setup.py with arguments is deprecated. There is no benefit in keeping this.
https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

.github/workflows/build-and-test.yml
fsimonis: Makes sense until we have a solution for the tests.

.github/workflows/build-and-test.yml
fsimonis: I removed the setup_test job above.

.github/workflows/build-spack.yml
fsimonis: Possible, but we don't even do this in precice.

.github/workflows/build-and-test.yml
BenjaminRodenberg: But we could do `pip install` instead to regularly test the installation procedure, no?

README.md
fsimonis: This is part of the setup.py deprecation. Should I split this?

.github/workflows/build-and-test.yml
BenjaminRodenberg: A right. That's defining the order of the individual jobs. Thanks!

README.md
BenjaminRodenberg: Yes. Let's do this in a different PR.

docs/ReleaseGuide.md
fsimonis: It will take substantially longer in the CI though.
A problem with cron jobs is that failure are not easy to discover.

.github/workflows/build-spack.yml
BenjaminRodenberg: I think it is/was generally useful to test this regularly. But it also comes with a considerable maintainment effort. So we can sacrifice this. Then there is the risk that when we release the spack package it does not build anymore. But I think if spack now uses pip there are much less possible sources of error. So we could probably skip this test.

spack/repo/packages/py-pyprecice/package.py
fsimonis: Not in the repo. It leads to duplication without benefit. Package updates should be easy enough using the fork of the maintainer.

docs/ReleaseGuide.md
BenjaminRodenberg: We could also solve this with a `workflow_dispatch` that you one can run when releasing.

spack/repo/packages/py-pyprecice/package.py
BenjaminRodenberg: I added the `package.py` here to regularly perform the tests. If we don't have the tests anymore we also don't need the (duplicated) `package.py` true.

.github/workflows/build-spack.yml
fsimonis: > Then there is the risk that when we release the spack package it does not build anymore.

Testing during the release process eliminates this issue.

docs/ReleaseGuide.md
BenjaminRodenberg: Testing spack is generally a bit a hen-and-egg-problem. If we remove the `package.py` here we could also consider the complete spack package as a downstream dependency: If it breaks after a release we need a bugfix. But that's ok and should not happen too often.

I'm optimistic that with spack relying on `pip` we will not have a lot of trouble in the future. If there is trouble we can still go back and add some test. My conclusion: Let's remove this optional step see how this will turn out.

.github/workflows/build-spack.yml
BenjaminRodenberg: See above. Maybe we can also take this risk. We are regularly testing the installation with `pip` and spack now basically uses `pip`. If we learn that this is not stable enough, we can still introduce additional test (automatized/during release -> let's see).

.github/workflows/build-and-test.yml
fsimonis: Restored

.github/workflows/build-and-test.yml
fsimonis: restored

.github/workflows/build-spack.yml
fsimonis: Restored

docs/ReleaseGuide.md
fsimonis: A good reason to manually test a release is that one needs to test the released package anyhow when adding the version to spack.
If the same person who tests during the release also updates the spack package later on, then there will be no rebuild of dependencies necessary.

fsimonis:
I updated this PR to only remove spack-specific things.

spack/repo/packages/py-pyprecice/package.py
BenjaminRodenberg, line 1: is there anything that we need to migrate from here to https://github.com/spack/spack/pull/45105 ?

BenjaminRodenberg (approved):
Looks good :+1: Let's :boom: the spack tests.

spack/repo/packages/py-pyprecice/package.py
fsimonis: Nope. This PR should have no impact on spack.

PR #223, opened 2024-10-10
Improve install error messages

fsimonis:
This PR adds a more helpful error message when the python bindings cannot find preCICE using pkg-config.
It also silences the NumPy warning to prevent it from distracting from the actual problem.


IshaanDesai (approved):
Looks good :+1:

PR #225, opened 2024-11-25
Add exception support

### LINKED ISSUE #173: Support preCICE Exception system
In https://github.com/precice/precice/pull/1594 we are currently introducing an exception system for preCICE. We should also support this in the python bindings.

Some necessary steps that I see here:

- [ ] Cython has some support for C++ Exception, but we need to explicitly state that certain functions may throw an exception. For details see https://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html#exceptions.
- [ ] Additionally, we can define our own exception system (basically map thrown C++ preCICE exceptions to python exceptions). We should use larger python Frameworks as a template here ([numpy, for example, also defines its own exceptions](https://github.com/numpy/numpy/blob/ea0b1708b27e683201285da53c6afda24bc8ba74/numpy/exceptions.py#L85))
- [ ] We should also make sure to be able to catch unknown exceptions (imagine C++ preCICE introduces a `FooError` upstream, then we want to at least raise a generic `precice.Error` here until we implement `precice.FooError`)
- [ ] We should also add (some) tests for error handling. I think we can simply extend our mocked version of preCICE with some dummy exceptions and explicitly test for them to make sure that passing the error generally works.

fsimonis:
This PR adds exception support for the entire API.

Note that both `const` and `except` are incompatible, hence I had to remove the `const` member specifiers from the API.
This shouldn't be a problem as `const` is purely decorative as there is no `const` usage in python.

This implementation translates all exceptions into `RuntimeError`s on the python side. We could provide a custom `Error` as part of the `precice` module to make this cleaner.

Closes #173

### DIFF
--- a/cyprecice/Participant.pxd
+++ b/cyprecice/Participant.pxd
@@ -11,8 +11,6 @@ cdef extern from "precice/Participant.hpp" namespace "precice":
 
         Participant (const string&, const string&, int, int, void*) except +
 
-        void configure (const string&)
-
         # steering methods
 
         void initialize ()
--- a/cyprecice/Participant.pxd
+++ b/cyprecice/Participant.pxd
@@ -13,17 +13,17 @@ cdef extern from "precice/Participant.hpp" namespace "precice":
 
         # steering methods
 
-        void initialize ()
+        void initialize () except +
 
-        void advance (double computedTimestepLength)
+        void advance (double computedTimestepLength) except +
 
         void finalize()
 
         # status queries
 
-        int getMeshDimensions(const string& meshName)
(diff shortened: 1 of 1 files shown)


BenjaminRodenberg (approved):
I manually tested how `Participant` and `get_mesh_vertices` behaves for faulty input and it looks good to me.

Ideally, we would also test for exceptions in the automated testing. But I do not really see how we can realize this without re-implementing the whole exception mechanism in the mocked preCICE. So let's not do it.

@fsimonis Feel free to merge.

PR #227, opened 2025-02-27
Use `venv` for CI

BenjaminRodenberg:
Required by recent pip versions (introduced with upgrade of Ubuntu base image for CI)


IshaanDesai (approved):
Makes sense :+1:

PR #228, opened 2025-02-27
Remove Dockerfile and release workflow

BenjaminRodenberg:
The Dockerfile that we use for building [`precice/python-bindings`](https://hub.docker.com/r/precice/python-bindings) currently breaks. The reason is a change in policy regarding installation of Python packages in recent pip versions (see [here](https://github.com/precice/python-bindings/actions/runs/13568176514/job/37926081519#step:8:690)).

@MakisH We can pass `--break-system-packages` to install the bindings system-wide. Does this meet your requirements for the systemtests? Or would you rather follow a different approach? Are you actually using the image `precice/python-bindings` or are you installing via `pip install git+https://github.com/precice/python-bindings.git@develop`?


MakisH:
This appears now because in https://github.com/precice/precice/pull/2200 we upgraded to Ubuntu 24.04, as decided in the last coding days.

The images published by this repository or by `precice/precice` are not used by the system tests.

Instead, we use [this Dockerfile](https://github.com/precice/tutorials/blob/develop/tools/tests/dockerfiles/ubuntu_2404/Dockerfile) and install in a venv:

https://github.com/precice/tutorials/blob/962793d5b173d828e2309af017076d8965fcfe79/tools/tests/dockerfiles/ubuntu_2404/Dockerfile#L85-L87

However, this is not even used by our tutorials at the moment, as each Python-based tutorial initializes each own venv. See https://github.com/precice/tutorials/issues/584.

The correct approach for this repository should not be the `--break-system-packages`, but using a virtual environment. This is what the new policy encourages.

BenjaminRodenberg:
I decided to remove the Dockerfile and the release workflow. I do not think we need it anymore and should use virtual environments instead.

MakisH (approved):

BenjaminRodenberg:
I took care of cleaning up issues related to the topic of releasing the python bindings via docker (see links above).

@fsimonis and @IshaanDesai putting you as reviewers here because of the many issues affected by this PR. I think removing the docker image in favor of pip and venv should save us a lot of maintainment work in the future. Please let me know if I'm missing something. I will merge this PR as soon as I have your approval.

IshaanDesai (approved):
Makes sense :+1:

PR #230, opened 2025-03-12
Update installation instructions

BenjaminRodenberg:
Using `setup.py install` is deprecated and since Ubuntu 24.04 usage of Python venvs is strongly encouraged. Inside of a virtual environment the `--user` option is not necessary.

We still need to cleanup some other issues related to dependency management, testing and the build process. Let's focus on some obvious problems in the documentation here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,13 +24,13 @@ Generally, it is recommended to work in a virtual environment when using the pre
 
 Before running any of the commands below, please create a virtual environment at a location of your choice by running:
 
-```
+```bash
 python3 -m venv .venv
 ```
 
 Using `.venv` is a common choice as path of your virtual environment. But you can use any path here. Next activate the virtual environment by running
 
-```
+```bash
 . .venv/bin/activate
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,13 +20,23 @@ Please refer to [the preCICE documentation](https://www.precice.org/installation
 
 ## Installing the package
 
-We recommend using pip3 (version 19.0.0 or newer required) for the sake of simplicity. You can check your pip3 version via `pip3 --version`. To update pip3, use the following line:
+Generally, it is recommended to work in a virtual environment when using the preCICE Python bindings. For this purpose, you might have to run `apt install python3-venv` first.
+
+Before running any of the commands below, please create a virtual environment at a location of your choice by running:
```
IshaanDesai, line 25:
```suggestion
Create a virtual environment in your working directory by running:
```
I would suggest sticking to imperative mood as we do in most of our documentation.

README.md
```diff
@@ -20,13 +20,23 @@ Please refer to [the preCICE documentation](https://www.precice.org/installation
...
+
+```bash
+python3 -m venv .venv
+```
+
+Using `.venv` is a common choice as path of your virtual environment. But you can use any path here. Next activate the virtual environment by running
```
IshaanDesai, line 31:
```suggestion
Using `.venv` is a common choice as path of your virtual environment. But you can use any path here. Activate the virtual environment by running
```

IshaanDesai (approved):
Looks good :+1: just two minor language-based suggestions from my side.

PR #234, opened 2025-04-01
Move from adding individual changelog entries to directly editing the main file

IshaanDesai:
This repository does not have too many contributions or parallel developments, so adding individual changelog entries is an overkill. This PR reverts to using one file for the changelog entries.


BenjaminRodenberg (approved):
Totally agree. Let's do it this way. Starting with individual files makes things unnecessarily complicated.

Issue #236, opened 2025-04-04
`map_and_read_data` does not flatten coordinate array before passing it to C++


r-walloner:
If I follow the example given in `help(precice.Participant.map_and_read_data)`, `map_and_read_data` throws the following error.
```
Traceback (most recent call last):
  File "/home/robin/uni/ba-repo/investigate_relativeReadTime/B/B.py", line 24, in <module>
    data = participant.map_and_read_data("A-Mesh", "A-Data", coordinates, 0)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "cyprecice/cyprecice.pyx", line 1000, in cyprecice.Participant.map_and_read_data
  File "<stringsource>", line 48, in vector.from_py.__pyx_convert_vector_from_py_double
TypeError: only length-1 arrays can be converted to Python scalars
```

The problematic line seems to be [this one](https://github.com/precice/python-bindings/blob/c28df2b64a504b276f8df518cd9cf443a406ef3f/cyprecice/cyprecice.pyx#L1000).
It seems to assume the `coordinates` are already flattened: `[x_0, y_0, x_1, y_1, ...]`, but they are not: `[(x_0, y_0), (x_1, y_1), ...]`.

r-walloner:
### Minimal Example to reproduce
A.py
```python
import precice

participant = precice.Participant("A", "../precice-config.xml", 0, 1)

mesh = [[0, 0]]
vertex_ids = participant.set_mesh_vertices("A-Mesh", mesh)
participant.initialize()

t = 0
dt = 1

while participant.is_coupling_ongoing():
    t = t + dt
    print("Timestep: ", t)

    data = t
    print("Writing data: ", data)
    participant.write_data("A-Mesh", "A-Data", vertex_ids, [data])

    participant.advance(dt)

participant.finalize()
```

B.py
```
import precice

participant = precice.Participant("B", "../precice-config.xml", 0, 1)

participant.set_mesh_access_region("A-Mesh", [-1, 1, -1, 1])
participant.initialize()

t = 0
dt = 1

while participant.is_coupling_ongoing():
    t = t + dt
    print("Timestep: ", t)

    relative_read_time = dt * 0
    print("Relative read time: ", relative_read_time)
    print("Absolute read time: ", t - dt + relative_read_time)

    coordinates = [(0.0, 0.0)]
[...]

PR #239, opened 2025-06-15
Modify docstrings to clearly state the flexibility of the API in accepting multidimensional data structures

IshaanDesai:
From the discussion in https://github.com/precice/python-bindings/issues/236, it emerged that there is confusion about the acceptable ways to provide multidimensional data structures to the API. The examples in the docstrings primarily show that NumPy arrays can be used, but the API also accepts a list of lists, or a list of tuples. This PR modifies the docstrings to clearly state this flexibility of the API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## latest
 
+* Modify docstrings to higlight flexbility of the API in terms of its ability to handle multidimensional input data structures https://github.com/precice/python-bindings/pull/239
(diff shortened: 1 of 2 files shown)


MakisH (approved):
This definitely helps.
The individual changes/examples make sense.

What about the `map_and_read_data`, which triggered the discussion? It also needs an example.

I have not checked if anything else needs a similar example, but since `set_mesh_vertices` is the first one to call, it would be kind of expected that the same applies to all.

IshaanDesai:
> What about the `map_and_read_data`, which triggered the discussion? It also needs an example.

I'll add an example here only.

PR #241, opened 2025-07-02
Renaming and using a newer workflow for publishing according to the trusted publishing in PyPI

IshaanDesai:
The new workflow needs to be tested with a release candidate.


.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+    name: Build package
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
+      - run: pip install --upgrade build
+      - name: Build package
+        run: pyproject-build
```
fsimonis, line 19: We need to build a source distribution:

```suggestion
        run: pyproject-build --sdist
```

fsimonis: The wheel build doesn't work without additional work (#242).
This situation is unchanged from the current publishing workflow though.

PR #242, opened 2025-07-02
Migrate to pyproject and tox

### LINKED ISSUE #152: Automatically get dependencies in the workflow to publish package on PyPI
Right now the dependencies of the bindings are manually specified in the [pythonpublish](https://github.com/precice/python-bindings/blob/develop/.github/workflows/pythonpublish.yml) workflow. This is problematic when a release contains the addition of a new dependency, as then there is no way to check if the dependencies in the publishing workflow are correct. One way to check this is to do a pre-release using a release candidate, but this is a cumbersome solution. A solution would be to have the workflow detect the dependencies automatically from the `pyproject.toml` file.

This issue is inspired by the release [v2.5.0.0](https://github.com/precice/python-bindings/releases/tag/v2.5.0.0) in which the added dependency of `pkgconfig` led to the PyPI publication workflow [to fail](https://github.com/precice/python-bindings/actions/runs/2956563566).

### LINKED ISSUE #208: Move from `setup.py` to `pyproject.toml`
`pyproject.toml` is the current standard for providing metadata and dependency information. Requiring setuptools>=61 (#207) is a step into this direction. See also https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html.

### LINKED ISSUE #214: Update tests to work with setuptools >= 72
Support for the `test` command has been removed in Setuptools 72. Currently, our (mock) tests require this command. We, therefore, need to update our tests to make them futureproof.

### LINKED ISSUE #217: Merge `setup.cfg` into `pyproject.toml`
`pyproject.toml` is the recommended format and we are mainly using the `setup.cfg` for versioneer. Versioneer also supports `pyproject.toml`. Therefore, moving should be possible. See https://github.com/python-versioneer/python-versioneer?tab=readme-ov-file#vendored-mode.

### LINKED ISSUE #222: Moving tests to tox
In tox, one can run disable the automatic installation of the package and run custom commands to setup an environment.

This should allow us to build the mock locally, then install the bindings and test against them.

https://tox.wiki/en/latest/config.html#skip_install

https://tox.wiki/en/latest/config.html#commands

We may need to LD_PRELOAD the mock though.

fsimonis:
This PR migrates from using setup.py and setup.cfg to using pyproject.toml.
It also replaces versioneer with setuptools-git-versioning.

The testing is now handled by tox, which sets an env used by setup.py to determine if it is in a testing environment.

Closes #222 
Closes #217 
Closes #208 
Closes #214 
Closes #152 

Note regarding #37: wheel build now with `pyproject-build -w`


pyproject.toml
```diff
@@ -1,3 +1,31 @@
...
-requires = ["setuptools>=61,<72", "wheel", "Cython>=0.29", "packaging", "pip>=19.0.0", "numpy", "mpi4py", "pkgconfig"]
+requires = ["setuptools>=61", "wheel", "Cython>=0.29", "packaging", "pip>=19.0.0", "numpy", "mpi4py", "pkgconfig", "setuptools-git-versioning"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "pyprecice"
+dynamic = [ "version" ]
+readme = "README.md"
+description="Python language bindings for the preCICE coupling library"
+license = "LGPL-3.0"
```
fsimonis, line 10: Is it LGPL-3.0 or LGPL-3.0-or-later?

pyproject.toml
```diff
@@ -1,3 +1,31 @@
...
+  "mpi4py", # mpi4py is only needed, if preCICE was compiled with MPI see https://github.com/precice/python-bindings/issues/8
+]
+
+authors = [
+  {name = "the preCICE developers", email = "info@precice.org"},
+]
+
+[project.urls]
+Repository = "https://github.com/precice/python-bindings.git"
+Issues = "https://github.com/precice/python-bindings/issues"
```
fsimonis, line 25: There are more urls supported here: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#urls

setup.py
```diff
@@ -67,13 +17,17 @@ def get_extensions(is_test):
...
     print("Found preCICE version " + pkgconfig.modversion('libprecice'))
 
+    compile_args = ["-std=c++17"]
+    link_args = []
+    include_dirs = [numpy.get_include()]
+    bindings_sources = ["cyprecice/cyprecice.pyx"]
     compile_args += pkgconfig.cflags('libprecice').split()
 
-    if not is_test:
+    if os.environ.get('TOX_ENV_NAME') is not None:
```
fsimonis, line 26: We currently use TOX_ENV_NAME to detect a testing scenario.
Maybe an env like `PYPRECICE_MOCKED` would be a better fit to simplify testing outside of tox environments.

.github/workflows/build-and-test.yml
```diff
@@ -65,16 +65,15 @@ jobs:
...
+        run: pip install tox
+      - name: Run tox
         env:
-          PKG_CONFIG_PATH: "precice-core/build"
           PKG_CONFIG_SYSTEM_INCLUDE_PATH: 1
         run: |
           export CFLAGS=-I$GITHUB_WORKSPACE
-          python3 setup.py test
+          export PKG_CONFIG_PATH=$(readlink -f "precice-core/build")
+          tox
```
fsimonis, line 76: We can probably replace a substantial part of the workflow with tox itself.

pyproject.toml
```diff
@@ -1,3 +1,31 @@
 [build-system]
-# PEP 518 - minimum build system requirements
-requires = ["setuptools>=61,<72", "wheel", "Cython>=0.29", "packaging", "pip>=19.0.0", "numpy", "mpi4py", "pkgconfig"]
+requires = ["setuptools>=61", "wheel", "Cython>=0.29", "packaging", "pip>=19.0.0", "numpy", "mpi4py", "pkgconfig", "setuptools-git-versioning"]
```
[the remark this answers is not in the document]
IshaanDesai: Shall we try removing pip and checking?

pyproject.toml
IshaanDesai: Unless we want to extend to apply the terms coming from LGPL versions > 3.0, we should stick to LGPL-3.0. @MakisH do you have an idea of what do we go for?

pyproject.toml
IshaanDesai: The Changelog URL seems like a good one to have.

pyproject.toml
IshaanDesai: Lets keep what we have now and change it later if required.

IshaanDesai (approved):
Thanks a lot for this!

PR #256, opened 2026-04-10
Fix passing MPI communicators Participant

### LINKED ISSUE #188: Add support for custom MPI communicator
The python bindings currently don't support the constructor which provides a custom MPI_Communicator to preCICE.

https://precice.org/doxygen/develop/classprecice_1_1Participant.html#a292d1a7e046ac3c083c3eec09bdb1a48

This locks preCICE into using MPI_COMM_WORLD and blocks reconstruction of the Participant.

Resulting issue: https://github.com/precice/precice/issues/828

fsimonis:
This PR fixes passing communicators to the participant initialization.

It used `mpi4py.MPI._addressof` to get the address to the underlying C comm and then forwards that to the participant interface.

Closes #188

### DIFF
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -9,7 +9,6 @@ cimport cyprecice
 cimport numpy
 import numpy as np
 from mpi4py import MPI
-
 import warnings
 from libcpp.string cimport string
 from libcpp.vector cimport vector
--- a/cyprecice/cyprecice.pyx
+++ b/cyprecice/cyprecice.pyx
@@ -9,6 +9,7 @@ cimport cyprecice
 cimport numpy
 import numpy as np
 from mpi4py import MPI
+
 import warnings
 from libcpp.string cimport string
 from libcpp.vector cimport vector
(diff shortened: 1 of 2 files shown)


valentinaschueller:
@fsimonis I can confirm that this works!

fsimonis:
I did some more digging and mpi4py has functionality to expose the address directly.
Meaning we can simply hotfix this.

cyprecice/cyprecice.pyx
```diff
@@ -9,6 +9,7 @@ cimport cyprecice
 cimport numpy
 import numpy as np
 from mpi4py import MPI
+
```
IshaanDesai:
```suggestion
```

IshaanDesai (approved):
Looks good 👍 @valentinaschueller thanks for confirming that this works! I hope the latest solution also works for you.

valentinaschueller:
It does! Thank you both for taking care of this so quickly 🙂

IshaanDesai:
> It does! Thank you both for taking care of this so quickly 🙂

Our pleasure. I will do a bugfix release soon.

